PR #1, opened 2015-08-21
Some initial revisions of the data model

no-reply:
I was charged with
  1. Replace `PrimarySourceSet:dplaer:timePeriod` with something from the period metadata schema
  2. Replace `PrimarySource:dplaer:aggregation` with something else - possibly `schema:hasPart`?
  3. Replace `TeachingGuide:dplaer:question` and `TeachingGuide:dplaer:activity` with a single field (`schema:text` or something better, like “body”). Expect that the field format will be markdown.  
  4. Possibly replace `PrimarySource:text`, and `PrimarySourceSet:text` with something better, like “body”.  Expect that their field format will be markdown.

Resolving each:
  1. Using `dct:temporal`, with the implication that the value is a `dct:PeriodOfTime`. I owe the Perio.do folks an email, and will get their opinion on applying that class to their `skos:Concept`s
  2. Using `dct:references`, which keeps the semantics here loose.
  3. This revision keeps `schema:text` (and therefore also `schema:Text`), but introduces the markdown datatype. The implication is that `:Markdown` is a subclass of `schema:Text`.
  4. It's okay to leave these as `schema:Text` unless we think we will be storing markdown in them. If we do think that, we can do the type coercion seen on `TeachingGuide.text`.

I've also loosened up the typing throughout, so `PrimarySource`s are just `MediaObject`s, and `TeachingGuide`s are just `CreativeWork`s. I think this is fine, and completes the process of internal namespace from the picture. We can always think about supporting a broader educational resources ontology that addresses these issues specifically if such a thing reveals itself.

### DIFF
--- a/etc/PrimarySourceSetDataModelSample.json
+++ b/etc/PrimarySourceSetDataModelSample.json
@@ -138,7 +138,7 @@
   ],
   "inLanguage": [
     {
-      "@id": "http://www.lexvo.org/page/term/eng/English",
+      "@id": "http://lexvo.org/id/term/eng/English",
       "@type": "Language",
       "name": "English"
     }
--- a/etc/PrimarySourceSetDataModelSample.json
+++ b/etc/PrimarySourceSetDataModelSample.json
@@ -3,13 +3,13 @@
     "http://schema.org/",
     {
       "ccss": "http://www.corestandards.org/",
-      "dc": "http://purl.org/dc/elements/1.1/",
+      "dct": "http://purl.org/dc/terms/",
       "dcmitype": "http://purl.org/dc/dcmitype/",
     }
   ],
-  "@id": "http://dp.la/primarysourceset/frenchindinanwar",
+  "@id": "http://dp.la/primarysourceset/frenchindianwar",
   "@type": "CreativeWork",
-  "dc:type": "dcmitype:InteractiveResource",
+  "dct:type": "dcmitype:InteractiveResource",
   "name": "The French and Indian War",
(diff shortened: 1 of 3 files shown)


etc/PrimarySourceSetDataModelSample.json
```diff
@@ -37,7 +37,7 @@
       "name": "Thirteen Colonies"
     }
   ],
-  "dplaer:timePeriod": [
+  "dct:temporal": [
     { "name": "Colonial America" }
```
AudreyAltman: Can we add a statement here asserting `@type: dct:PeriodOfTime`?  Also, I just want to confirm that `dct:PeriodOfTime` is expected to have a `name` property.

etc/PrimarySourceSetDataModelSample.json
```diff
@@ -138,7 +138,7 @@
   ],
   "inLanguage": [
     {
-      "@id": "http://www.lexvo.org/page/term/eng/English",
+      "@id": "http://lexvo.org/id/term/eng/English",
```
AudreyAltman: Good catch!

AudreyAltman: In response to point 4. above:

> It's okay to leave these as schema:Text unless we think we will be storing markdown in them. If we do think that, we can do the type coercion seen on TeachingGuide.text.

We should plan on both `PrimarySource:text` and `PrimarySourceSet:text` being markdown, so I say go ahead and use the same type coercion seen on `TeachingGuide.text`.

PR #8, opened 2015-08-27
Feature/primary source set

AudreyAltman:
This PR introduces a basic CRUD interface for primary source sets.

Additional changes:
Get set up for rspec testing.
Remove assets group from Gemfile b/c it is not supported in more recent versions of Rails
Remove some unnecessary directories and files.

Still to do (in subsequent PRs):
Add media upload capability.
Add CRUD interfaces for other models.


no-reply:
Since a couple of the changes in this PR are now mine, I want to throw it back to @AudreyAltman for a :+1: before merging.

I'm aware that there were some problems with the automation side of this, and it wasn't clear to me whether the fix would be on this end or there. Either way, this is now passing CI, so I think we should consider it in a healthy state. If there's something wrong with the setup, it can be handled in another PR.

AudreyAltman:
:+1: I think this is good to go.  I will :squash:

AudreyAltman:
Okay, I :squash:ed to 3 commits.  Ready to :merge: @no-reply approves.

no-reply:
:rocket:

PR #9, opened 2015-08-30
Feature/extended crud

AudreyAltman:
This PR introduces CRUD interfaces for `source`, `guide`, and `author`.  

`source` and `guide` are very similar - they both have `belongs_to` relationships to `source_set`, and their routes are both nested under `source_set`.  Neither have their own `index` route or view; instead, they are listed on their parent `source_set#show` view.

`author` has a `has_many_and_belongs_to` relationship with both `source_set` and `guide` and as such, required the creation of additional associations tables.

This PR also introduces the [redcarpet](https://github.com/vmg/redcarpet) gem to render Markdown.  Five of the data entry fields accept Markdown, so it needs to be rendered to the user in the `show` views so that they can verify that their data entry was correct.  Look at the [render markdown](https://github.com/dpla/primary-source-sets/commit/7ab719de95d06ce61d6dd0d07290984ef4925209) commit to see only the changes having to do with rendering Markdown.

The best way to test the views is to play with them in your browser.  Remember to do `bundle install` and  `rake db:migrate`.

If you want to run the automated tests locally, remember to do `bundle install` and `rake db:test:prepare`.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,12 @@
 module ApplicationHelper
+
+  ##
+  # @param String text A markdown formatted String
+  # @return String An HTML representation of the markdown string given in 'text'
+  # @raise TypeError if @param is not a String
+  def markdown(text)
+    options = { autolink: true, footnotes: true }
+    Redcarpet::Markdown.new(Redcarpet::Render::HTML, options).render(text)
+      .html_safe
+  end
 end
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -1,0 +1,8 @@
+class Source < ActiveRecord::Base
+  belongs_to :source_set
+  validates :aggregation, presence: true
+
+  def aggregation_uri
+    Settings.frontend.url + 'item/' + aggregation
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem 'sass-rails', '~> 4.0.3'
 gem 'jquery-rails'
 gem 'turbolinks'
 gem 'config', '~> 1.0.0'
+gem 'redcarpet', '~> 3.3'
 
 group :test, :development do
   gem 'rspec-core', '~>3.1'
--- a/app/controllers/authors_controller.rb
+++ b/app/controllers/authors_controller.rb
@@ -1,0 +1,54 @@
+##
+# Handles HTTP requests for Authors
+#
+# @see Author
+class AuthorsController < ApplicationController
+  def index
+    @authors = Author.all
+  end
+
+  def show
+    @author = Author.find(params[:id])
+  end
+
+  def new
+    @author = Author.new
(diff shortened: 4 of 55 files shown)


app/models/source.rb
```diff
@@ -0,0 +1,8 @@
+class Source < ActiveRecord::Base
+  belongs_to :source_set
+  validates :aggregation, presence: true
+
+  def aggregation_uri
+    'http://dp.la/item/' + aggregation
```
anarchivist: Can we make this URI base configurable? This will need to be different for staging vs. production.

AudreyAltman: Excellent idea.  Do you think we could make this its own PR?  I am asking b/c there is another PR pending (#12) that adds the `config` gem.  Once that gem is in place, making this configurable will be much easier.  The other option would be to wait until PR#12 is merged and then rebase this branch.

AudreyAltman: Since PR#12 has been merged, I rebased to develop used the `config` gem to make this URI configurable.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,7 @@
 module ApplicationHelper
+  def markdown(text)
```
no-reply: Docs?

I'm not inclined to be picky about controllers, but I think a docstring on helper methods is a good idea.

no-reply: A unit test would be good too. :)

AudreyAltman: @no-reply can you tell me more about your thoughts on this? I'm not sure I understand what you are suggesting.

no-reply:
This looks good to me. Apart from the comments about the helper method docs/tests, I think it's ready to go.

app/helpers/application_helper.rb
no-reply: Something like:
```ruby
##
# @param [{what types does Markdown#render accept?}] text  A markdown formatted string {(or IO object?)}
# @return [String] An HTML representation of the markdown string given in `text`
# @raise {???? does Redcarpet ever throw errors on #render or #html_safe?}
```

And a test that passes in a simple markdown string and checks that the output looks at least vaguely right; maybe also tests for any expected error conditions.

AudreyAltman: That makes sense, thanks, and good suggestions!

anarchivist:
I'm in agreement with @no-reply. :rocket:

app/helpers/application_helper.rb
AudreyAltman: @no-reply I couldn't find any specific exceptions from Redcarpet, but I added some comments to the helper method, added a test, and added a check for String input:  https://github.com/dpla/primary-source-sets/commit/aeefa533ae04b851d135e3a3609d073449aa7363

AudreyAltman:
:squash:ed

PR #10, opened 2015-08-30
update json samples

AudreyAltman:
This simplifies the data model and brings it into alignment with the relational database.


no-reply:
No need for locations? :+1:

As long as we can count on, e.g. "Georgia" never meaning anything more than a set of unicode characters in a name, then this looks good to me.

AudreyAltman:
I don't think we need locations.  There's no plan to sort on them or anything - it's just, as you say, unicode characters.

PR #11, opened 2015-09-01
append pss/ to routes

AudreyAltman:
This appends `pss/` to all routes.


AudreyAltman:
DO NOT MERGE.  If this solves the problem with automation, I will need to correct the rspec tests before merging.

PR #12, opened 2015-09-01
append primary-source-sets/ to routes

AudreyAltman:
This appends `primary-source-sets/` to routes.  It also introduces `config` gem to manage settings.


AudreyAltman:
:+1:

PR #13, opened 2015-09-01
Add build-essentials to provisioned packages

no-reply:
Building Ruby could/did(?) fail, because `make` was not installed. This
installs the dependencies for a Ruby build.


AudreyAltman:
:+1:

PR #15, opened 2015-09-04
add set slugs with FriendlyId

AudreyAltman:
This adds slugs to SourceSets with the [FriendlyId gem](https://github.com/norman/friendly_id).

The FriendlyId gem allows you to use human-readable ids in URLs, which is good for usability and SEO.  For example, instead of:

`http://dp.la/primary-source-sets/sets/82364`

you would get something like this:

`http://dp.la/primary-source-sets/sets/french-and-indian-war`

In this configuration, the name field for SourceSets is used to automatically generate the slug.

### DIFF
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -1,0 +1,88 @@
+# FriendlyId Global Configuration
+#
+# Use this to set up shared configuration options for your entire application.
+# Any of the configuration options shown here can also be applied to single
+# models by passing arguments to the `friendly_id` class method or defining
+# methods in your model.
+#
+# To learn more, check out the guide:
+#
+# http://norman.github.io/friendly_id/file.Guide.html
+
+FriendlyId.defaults do |config|
+  # ## Reserved Words
+  #
+  # Some words could conflict with Rails's routes when used as slugs, or are
+  # undesirable to allow as slugs. Edit this list as needed for your app.
+  config.use :reserved
+
+  config.reserved_words = %w(new edit index session login logout users admin
+    stylesheets assets javascripts images)
+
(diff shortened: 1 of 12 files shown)


config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,88 @@
...
+  # calls to `Module.extend`.
+  #
+  # config.use :finders
+  #
+  # ## Slugs
+  #
+  # Most applications will use the :slugged module everywhere. If you wish
+  # to do so, uncomment the following line.
+  #
+  # config.use :slugged
```
no-reply: Would it be a good idea to uncomment this, rather than configuring on a per module basis at, e.g. https://github.com/dpla/primary-source-sets/pull/15/files#diff-8aea8a5bfdb4698272ac5b9ee26ba88fR7 ?

AudreyAltman: @no-reply we are only planning to use friendly ids for the SourceSet model, and not for any other models.

no-reply:
@AudreyAltman can you say a little more, perhaps in the commit message, about the FriendlyId gem?  A little info about how this is different from finding by name and some relevant links would be good to have there for reference.

AudreyAltman:
@no-reply good suggestion, I'll add a comment to the model.

config/initializers/friendly_id.rb
no-reply: ah :+1:. It's probably more clear this way, then.

AudreyAltman:
@no-reply okay, comments added to model

no-reply:
:100: I'll merge this unless @anarchivist wants to take a look.

PR #16, opened 2015-09-08
authorization with devise gem

AudreyAltman:
This feature adds basic authorization to the app with the [Devise gem](https://github.com/plataformatec/devise).

Authentication is needed for this app so that admins (ie. DPLA staff) can log in to perform CRUD actions in the database.  After our initial launch in October, non-logged-in users will be able to view sets, sources, and guides; but until launch, only admins should have the ability to access any part of the application.  So for now, I added `before_action :authenticate_admin!` to the `ApplicationController`.  When we launch in October, this will need to change.

I kept most of the default Devise settings, expressed in the database migration, `config/initializers/devise.rb` and `config/locales/devise.en.yml`.  One exception is that I added the DPLA email address in [line 13](`config/initializers/devise.rb`) of `config/initializers/devise.rb`.

The setting for the `Admin` model are very simple.  I am double-checking with Franky, but I think we only need one admin account that DPLA staff members can share, and its easy to make a new account from the command line.  So there is log-in and log-out capability from the app interfaces, but there is NOT ability to make a new account or reset a password.  We can always add these things in later if we decide we need them.

I made some spec settings changes to work with Devise: see `spec/support/devise.rb` and `spec/support/controller_macros.rb` and the new factory, `spec/factories/admins.rb`.  Although there appears to be a lot of changes to the existing specs, it's really just nesting all of the existing tests within a `context` block that logs in the admin.

I was also getting some errors with the rspec that I fixed by adding `.rspec` and `spec/rails_helper.rb`.  These files are auto-generated with you do `rails generate rspec:install`, which up to now I had neglected to do.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'turbolinks', '~> 2.5.3'
 gem 'config', '~> 1.0.0'
 gem 'redcarpet', '~> 3.3'
 gem 'friendly_id', '~> 5.1.0'
+gem 'devise', '3.4.1'
 
 group :test, :development do
   gem 'rspec-core', '~> 3.3.2'
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,7 +3,13 @@
 # Information in this file will be PUBLIC (this file is tracked in git)
 # Declarations in config/settings.local.yml override declarations in this file
 
+devise: 
+  mailer_sender: email@example.org
+
 frontend:
   # With trailing slash
-  url: http://dp.la/
+  url: http://example.com
+
+host: 'localhost'
+
 relative_url_root: /primary-source-sets
--- a/spec/controllers/authors_controller_spec.rb
+++ b/spec/controllers/authors_controller_spec.rb
@@ -1,111 +1,119 @@
-require 'spec_helper'
+require 'rails_helper'
 
 describe AuthorsController, type: :controller do
 
-  let(:author) { FactoryGirl. create(:author_factory) }
+  let(:author) { FactoryGirl.create(:author_factory) }
 
-  describe '#index' do
-
-    it 'set @authors variable' do
-      get :index
-      expect(assigns(:authors)).to eq([author])
-    end
-
-    it 'renders the :index view' do
-      get :index
-      expect(response).to render_template :index
-    end
+  it_behaves_like 'admin-only route', :index, :show, :edit, :new do
+    let (:resource) { FactoryGirl.create(:author_factory) }
   end
 
-  describe '#show' do
+  context 'admin logged in' do
+    login_admin
 
-    it 'sets @author variable' do
-      get :show, id: author.id
-      expect(assigns(:author)).to eq(author)
+    describe '#index' do
+
+      it 'set @authors variable' do
+        get :index
+        expect(assigns(:authors)).to eq([author])
+      end
+
+      it 'renders the :index view' do
+        get :index
+        expect(response).to render_template :index
+      end
     end
 
-    it 'renders the :show view' do
-      get :show, id: author.id
-      expect(response).to render_template :show
-    end
-  end
+    describe '#show' do
 
-  describe '#create' do
+      it 'sets @author variable' do
+        get :show, id: author.id
+        expect(assigns(:author)).to eq(author)
+      end
 
-    it 'creates a new author' do
-      expect do
-        post :create, author: attributes_for(:author_factory)
-      end.to change(Author, :count).by(1)
-    end
-
-    it 'redirects to the new author' do
-      post :create, author: attributes_for(:author_factory)
-      expect(response).to redirect_to Author.last
+      it 'renders the :show view' do
+        get :show, id: author.id
+        expect(response).to render_template :show
+      end
     end
 
-    context 'with invalid attributes' do
+    describe '#create' do
 
-      it 'does not save new author' do
+      it 'creates a new author' do
         expect do
+          post :create, author: attributes_for(:author_factory)
+        end.to change(Author, :count).by(1)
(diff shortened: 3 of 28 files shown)


spec/controllers/authors_controller_spec.rb
```diff
@@ -4,108 +4,112 @@
 
   let(:author) { FactoryGirl. create(:author_factory) }
 
-  describe '#index' do
+  context 'admin logged in' do
```
no-reply: Any need for tests without an active admin session?

AudreyAltman: @no-reply good suggestion, I'm on it.

no-reply:
With the introduction of `rails_helper` there's some overlap with `spec_helper` that we might want to clean up. Mainly:

  - We probably want to be requiring `rails_helper` rather than `spec_helper` in MVC tests; and
  - remove Rails (and Factory Girl) specific setup from `spec_helper` https://github.com/dpla/primary-source-sets/blob/feature/authorization/spec/spec_helper.rb

AudreyAltman:
I confirmed with FA that we only need one admin account that can be shared by all DPLA staff.

config/settings.yml
```diff
@@ -3,7 +3,10 @@
 # Information in this file will be PUBLIC (this file is tracked in git)
 # Declarations in config/settings.local.yml override declarations in this file
 
+email: info@dp.la
```
markbreedlove: Since this will be `mailer_sender`, e.g. in [`config/initializers/devise.rb`](https://github.com/dpla/primary-source-sets/blob/feature/authorization/config/initializers/devise.rb#L13), could it be named as such here? I could picture reading this without knowing how it's used, and thinking this is a recipient address.  It's also used specifically as a Devise-related setting, so maybe it should be renamed or categorized to make that clear (put under a `devise` or `authentication` element, for instance).

Also, if it's changed to something `@example.org`, it'll be a little safer to have in here as a default value.  (It's supposed to be overridden in `settings.local.yml`, as the comment suggests.

config/settings.yml
```diff
@@ -3,7 +3,10 @@
...
+
 frontend:
   # With trailing slash
   url: http://dp.la/
```
markbreedlove: Same here, regarding default values checked into git.

markbreedlove:
I recall, from adding Devise to Krikri months ago, that adding roles is a significant extra step, that we deferred to later with that project.  I'm generally very much against having one account of any kind shared among a group of users, especially an admin account.  Could the installation be set up so that individual users have their own accounts, with the responsibility that goes along with it of managing and not sharing their individual passwords?  As long as we're temporarily using `before_action :authenticate_admin!` it will still let them in.  This isn't so much a code suggestion as a deployment and support suggestion. :-)

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'turbolinks'
 gem 'config', '~> 1.0.0'
 gem 'redcarpet', '~> 3.3'
 gem 'friendly_id', '~> 5.1.0'
+gem 'devise', '3.4.1'
```
markbreedlove: It's good to see that this is pinned to a specific version, because I think we needed to do that with Krikri.

config/settings.yml
AudreyAltman: :+1:

markbreedlove:
Hi, @AudreyAltman:  Can you describe the changes made to the tests?  Were you just adding the 'admin logged in' contexts?

markbreedlove:
Thinking again about my comment about authentication and considering the time it would take to add create-an-account and change-password functionality.  That's a good point.  I think that maybe, though, considering that only two or a small handful people are going to be adding content to begin with, their accounts could be created manually and then barebones password-reset and add-user pages could be created in the next few weeks.

AudreyAltman:
@markbreedlove re:tests I was just adding the 'admin logged in' contexts to the existing tests, but per @no-reply's suggestion, I am also in the process of adding some tests for non-logged-in users.

AudreyAltman:
@markbreedlove re: create-an-account and change-password functionality - I am total agreement with you that password-reset and add-user pages should be added as time allows.  I would normally be in favor of adding them at the offset, but considering our tight timeline, I think it is okay to wait.

AudreyAltman:
I made edits based on comments from @no-reply and @markbreedlove: https://github.com/dpla/primary-source-sets/commit/04a9eee78157aa9a65b95b94f6e7bb6869ded4d3

1. I realized that `devise` was in the Gemfile twice(!) so I deleted one of the two statements.
2. I changed the settings in the `yml` files according to @markbreedlove's suggestions.
3. I attempted to move Rails specific setup from `spec_helper` to `rails_helper` and deleted redundancies - however I am not really clear on which setups are Rails-specific, so I would love some feedback about that.
4. I required `rails_helper` instead of `spec_helper` in all MVC tests.
5. I added a new `shared_example` in `rspec` to test that non-admin users get redirected to the admin login page when they try to access admin-only routes: https://github.com/dpla/primary-source-sets/blob/04a9eee78157aa9a65b95b94f6e7bb6869ded4d3/spec/support/shared_examples/admin_only_route.rb

no-reply:
This looks ready to me.

I'll look into the merge conflict, it's probably just line ordering in the Gemfile.

no-reply:
Fixed and (force) pushed. I'm ready to merge this, when it passes.

AudreyAltman:
@no-reply I noticed that the `devise` gem managed to sneak back into the `Gemfile` twice, so I fixed it.  I will :squash:.

no-reply:
:+1:

markbreedlove:
Looks good to merge.

Issue #2, opened 2015-08-21
Lint JSON-LD


no-reply:
Add a script to parse the example JSON-LD and verify that it produces the correct graph.

no-reply:
This was closed by #17.

PR #17, opened 2015-09-09
temporal, citation, and provenance

AudreyAltman:
This adds citation and provenance metadata to the `Source` class, and adds more detailed temporal metadata.  These changes are motivated by conversations I've had with FA.

Although the citation and provenance metadata is accessible by linking to the ore:Aggregation, we want to include it here so that if the ore:Aggregation disappears, as they sometimes do, we can retain the information here.

FA has decided that for the primary source sets time periods, she is going to use the [US History Content Standards eras](http://www.nchs.ucla.edu/history-standards/us-history-content-standards). Since we now have time periods with start and end dates, I took a stab at representing them using DC's PeriodOfTime class. 

A particular concern is this: As far as I know, the US History Content Standards do not have an RDF expression (maybe worth double-checking).  So, the URL that I used for @id is not any sort of official RDF URI, it's just the web page where the info exists.  Is that a legit way of representing that this time period term comes from a standardized vocabulary?  Is there a better way?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,4 +29,6 @@ group :test, :development do
   gem 'awesome_print', '~> 1.2.0'
   gem 'codeclimate-test-reporter', '~> 0.4.7', require: false
   gem 'database_cleaner', '~> 1.3.0', require: false
+
+  gem 'json-ld', '~>1.1.9'
 end
--- a/etc/PrimarySourceSetDataModelSample.json
+++ b/etc/PrimarySourceSetDataModelSample.json
@@ -43,7 +43,13 @@
     }
   ],
   "dct:temporal": [
-    { "name": "Colonial America" }
+    { 
+      "@id": "ushist:united-states-era-2",
+      "@type": "dct:PeriodOfTime",
+      "name": "Colonization & Settlement (1585-1763)",
+      "start": "1585",
+      "end": "1763",
+    }
   ],
   "description": "This collection uses primary sources to explore the French and Indian War.",
   "author": [
--- a/etc/PrimarySourceSetDataModelSample.json
+++ b/etc/PrimarySourceSetDataModelSample.json
@@ -5,6 +5,7 @@
       "ccss": "http://www.corestandards.org/",
       "dct": "http://purl.org/dc/terms/",
       "dcmitype": "http://purl.org/dc/dcmitype/",
+      "edm": "http://www.europeana.eu/schemas/edm",
       "text":
       {
         "@id": "http://schema.org/text",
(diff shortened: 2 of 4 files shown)


etc/PrimarySourceSetDataModelSample.json
```diff
@@ -43,7 +43,13 @@
     }
   ],
   "dct:temporal": [
-    { "name": "Colonial America" }
+    { 
+      "@id": "ushist:united-states-era-2",
+      "@type": "dct:PeriodOfTime",
```
no-reply: Probably we should use `edm:TimeSpan` (which is a subclass of `dct:PeriodOfTime`) for consistency.

Are `start` and `end` in some namespace? I'm missing the in Schema.org. Do we want `edm:begin/end`?

no-reply:
I'd like to put this through a JSON-LD parser and spot check. Can probably get to that tomorrow morning.

Will try to close #2 in the process.

etc/PrimarySourceSetDataModelSample.json
AudreyAltman: :+1: to `edm:TimeSpan` 
`start` and `end` were from DCMI, I just forgot the prefix (http://dublincore.org/documents/dcmi-period/ for reference).  But I think that `edm:begin/end` would be better.

no-reply: > But I think that edm:begin/end would be better.

:+1: 

`start/end` are referred to in that document as XML attributes. They aren't defined in DC Elements or Terms namespaces. Does make me curious what DCMI's guidance would be on this front, though.

no-reply:
I added a JSON-LD linter to the specs on this branch. One of the 

@AudreyAltman do you mind if I push to the branch so we can resolve this here? Or should I push a separate branch and add comments there?

AudreyAltman:
@no-reply go ahead and push to the branch!

no-reply:
With 77ee335, this closes #2.

no-reply:
The fix for the JSON turned out to be very small. The linter should make it easy to catch this in the future.

I think we should merge this as is, and create a task in Redmine to double check the RDF output, and check that it matches up with the actual application data.

Gemfile
```diff
@@ -23,6 +23,8 @@ group :test, :development do
   gem 'awesome_print'
   gem 'codeclimate-test-reporter', require: false
   gem 'database_cleaner', '~> 1.3.0', require: false
+
+  gem 'json-ld'
```
AudreyAltman: @no-reply can we tie this gem to a version number?

no-reply: can do.

AudreyAltman:
The linter looks great.  I made one small suggestion, but otherwise I think this is ready to :squash: and :merge:

no-reply:
I did the rebase, on this one, and have a code in two of the commits, perhaps @anarchivist or @markbreedlove should merge? :)

markbreedlove:
Sorry for the delay ...

PR #18, opened 2015-09-10
tighen gem dependencies

AudreyAltman:
This introduces versions for all gems.  It changes the dependency on `rails` from `4.1.6` to `~> 4.1.6`.  Whenever it does not conflict with the current `Gemfile.lock`, versions of gems match either the `heidrun` or `frontend` app.


no-reply:
:+1:

PR #19, opened 2015-09-11
add markdown fields

AudreyAltman:
This introduces markdown formatting to three new fields: `source_set.name`, `source.name`, and `guide.name`.

These three markdown fields differ from the previous markdown fields in that they need to be rendered as inline HTML elements.  Since `Redcarpet` by default encloses all markdown in `<p>` tags, I made a new helper method, `inline_markdown`.  Most of the changes to the views are simply using this new method to render `source_set.name`, `source.name`, and `guide.name`.

In addition, I added instructional text to the `source_set`, `guide`, and `source` input forms, and fixed a few typos.

This addresses [#7985](https://issues.dp.la/issues/7985)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -10,7 +10,37 @@ module ApplicationHelper
       .html_safe
   end
 
+  ##
+  # @param String text A markdown formatted String
+  # @return String An HTML representation of the markdown string given in 'text'
+  # without enclosing <p> tags.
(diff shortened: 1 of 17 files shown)


no-reply:
This looks good! :+1:

Needs a rebase on the current develop to merge.

AudreyAltman:
rebased and ready to :merge: (when tests pass)

PR #20, opened 2015-09-11
Add credits, citation to sources

AudreyAltman:
This introduces the ability to persists source citations and credit in the database, and enter them via the admin web interface.

It is important to persist these (rather than generating them dynamically through the API) because if an item disappears from the DPLA library, we need to be able to retain citation and credit information.  

This addresses [#7991](https://issues.dp.la/issues/7991)


no-reply:
Awesome. 

image

no-reply:
This also needs a rebase.

AudreyAltman:
rebased and ready to :merge:

PR #21, opened 2015-09-13
load view specs and require rails_helper

AudreyAltman:
This fixes view specs, which were failing.  We didn't detect that they were failing because they were not loading with the rest of the tests when running `rspec`.

To get the view specs to pass, I changed `require spec_helper` to `require rails_helper`.  This change was made to the rest of the specs in PR#16, but I had overlooked the views specs. 

To get the view specs to load, I added a statement to `spec_helper`.


no-reply:
:+1: rebase?

AudreyAltman:
rebased and ready to :merge:

PR #22, opened 2015-09-13
clean up gemfile

AudreyAltman:
This removes one of the two identical statements in the `Gemfile` for the `devise` gem.  It reorganizes the file so that similar mistakes will be less likely in the future.  It also adds the `rubocop` gem.


no-reply:
:+1: rebase?

PR #23, opened 2015-09-14
Fix/guide source routes

AudreyAltman:
This fixes `guide` and `source` routes so that they are shallowly nested under `source_set`.  The `shallow` option nests only `:index`, `:new`, and `:create` under `source_set`.  

Shallow nesting makes the routes simpler. For example, without shallow nesting, a route for a source might be `primary-source-sets/sets/french-and-indian-war/sources/52`.  With shallow nesting, the route would be `primary-soruce-sets/sources/52`, but the source's relationship to its parent set is maintained.

Shallow nesting will also make development for images and assets easier.

Shallow nesting changes the way you need to assert route helper methods, such as `redirect_to` and `guide_path` which accounts for a large portion of the changes in this PR.

This PR also adds a friendly id (ie slug) for guides.  Since shallowly-nested `guide` routes will no longer contain the name of their parent primary source set, for SEO purposes it will be good to have descriptive, human-readable slugs instead of IDs in the `guide` URLs.

This PR also adds an `index` route action to `guide` and `source`.  Previously, there were no `index` route actions; now, there are, and they redirect to the parent `source_set`.  It is a small improvement for usability and adhering to expected patterns.

Finally, this adds a helper method `source_name`, a minor refactor to help DRY out the views.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -2,11 +2,15 @@
 # Handles HTTP requests for guides
 #
 # @see Guide
-class GuidesController < ApplicationController
-  before_filter :load_source_set
+class GuidesController < ApplicationController 
+  before_filter :load_source_set, only: [:index, :new, :create]
+
+  def index
+    redirect_to @source_set
+  end
 
   def show
-    @guide = @source_set.guides.find(params[:id])
+    @guide = Guide.friendly.find(params[:id])
   end
 
   def new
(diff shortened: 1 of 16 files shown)


no-reply:
This looks straight-forward and good to me.  :+1: 

image

markbreedlove:
Looks good to me.  Do you want to squash the commits, or keep them as they are?

AudreyAltman:
I will :squash:  Also, just realized I should do a little more cleanup in the specs.  I'll both those things quickly.

AudreyAltman:
In the controller specs, I removed route params that are no longer necessary.  https://github.com/dpla/primary-source-sets/commit/c7656dd4e9f70748e32e377c8569d698b9971695

AudreyAltman:
:squash:ed

PR #24, opened 2015-09-28
Feature/media scaffolding

AudreyAltman:
This introduces the basic database tables, models, views, and controllers for media objects, including their relationships to source sets and sources.  It addresses [#8025](https://issues.dp.la/issues/8025) and follows the [ER Diagram](https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Relational+Database+Model).  This does NOT allow admins to upload actual media files, just their related metadata.

Media objects (images, video, audio, and documents) form a conceptual group of "assets".  Assets belong to "attachable" objects (sources or source sets).  "Assets" and "attachables" don't lend themselves to explicit representation in the relational database, nor did it seem helpful to make "asset" and "attachable" classes or mixins in the application.  Instead, "assets" and "attachables" are defined through relationships and accessors in the models.  These conceptual object types are helpful because they allow you to refer generically to, say, a source's `asset` without needed to know whether that `assets` is an image, video, etc.  There is a `shared_examples` [spec for "assets"](https://github.com/dpla/primary-source-sets/blob/980d72f8664f9ee84645b7a07a0c1ce39a82438c/spec/support/shared_examples/media_asset.rb) that describes the expected behavior of an "asset".

This refactors some controller specs, introducing some `shared_examples` to DRY them out.

To run tests locally:  remember to do `rake db:test:prepare`.

To run the server locally:  remember to clean up your database to get rid of any old data that conforms to previous database schema.  I find it easiest to do `bundle exec rake db:drop db:create db:migrate`.

### DIFF
--- a/app/controllers/audios_controller.rb
+++ b/app/controllers/audios_controller.rb
@@ -1,0 +1,65 @@
+##
+# Handles HTTP requests for Audios
+#
+# @see Audios
+class AudiosController < ApplicationController
+  before_filter :load_source, only: [:index, :new, :create]
+
+  def index
+    redirect_to @source
+  end
+
+  def show
+    @audio = Audio.find(params[:id])
+  end
+
+  def new
+    @audio = @source.build_audio
+  end
+
+  def edit
+    @audio = Audio.find(params[:id])
+  end
+
+  def create
+    @audio = @source.build_audio(audio_params)
+
+    if @audio.save
+      redirect_to @audio
+    else
+      render 'new'
+    end
+  end
+
+  def update
+    @audio = Audio.find(params[:id])
+
+    if @audio.update(audio_params)
+      redirect_to @audio
+    else
+      render 'edit'
+    end
+  end
+
+  def destroy
+    @audio = Audio.find(params[:id])
+    @audio.destroy
+
+    redirect_to @audio.source
+  end
+
+  private
+
+  def audio_params
+    params.require(:audio).permit(:source_id,
+                                  :mime_type,
+                                  :file_base)
+  end
+
+  ##
+  # Find the source through the HTTP route.
+  # This method is only for use those actions nested under source.
+  def load_source
+    @source = Source.find(params[:source_id])
+  end
+end
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -1,0 +1,65 @@
+##
+# Handles HTTP requests for Documents
+#
+# @see Document
+class DocumentsController < ApplicationController
+  before_filter :load_source, only: [:index, :new, :create]
+
+  def index
+    redirect_to @source
+  end
+
+  def show
+    @document = Document.find(params[:id])
+  end
+
+  def new
+    @document = @source.build_document
+  end
+
+  def edit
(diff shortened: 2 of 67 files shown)


markbreedlove:
I'm not sure what metadata are needed yet for the assets.  Some tables might benefit from a `meta` column, of type `text`, that we can use for storing JSON for properties and other encoding-related data.

Each audio or video asset will have derivatives.  For example, a video will have a H.264 (.mp4) file and a Webm (.webm) file for cross-browser compatibility.  There should probably be separate tables for those, but I'm not sure yet.  I can add those later.  I have to figure out what's most efficient for the transcoding process.  It depends on whether we handle transcoding internally as separate jobs for each derivative.  Each derivative will be created as a separate job if we use Amazon Elastic Transcoder, and we would probably benefit from multiple records for those, but I'm uncertain.

markbreedlove:
Regarding the `@attachable` properties and `attachable` methods:

Can you explain the advantage of using these properties versus just `<Model>.source`?

I'm not clear as to what `ApplicationController#load_attachable` provides because it seems that, for any given thing like an audios controller or a videos controller it should always know what kind of object the audio or video belongs to: e.g. a `Source`. I know that `Image`s can belong to `Source`s or `SourceSet`s, but maybe there should be two different models for what are different kinds of images, or maybe there needs to be some kind of one-to-many relationship.

There is also [this line](https://github.com/dpla/primary-source-sets/pull/24/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR18) in `ApplicationController` where it appears to be instantiating an object based on a string parameter received in the querystring and I'm wondering if that's safe.

AudreyAltman:
@markbreedlove - in response to your comments above, I've made some [changes].(https://github.com/dpla/primary-source-sets/commit/e2641df6accd85673cb9a4b66aafdb2b457c9dbe)

I think your instinct is correct that it would be simpler to refer to `attachable`s as `<Model>.source`.  I added an `Image.source` method and did away with `Document.attachable`, `Video.attachable`, and `Audio.attachable`.

I retained `Image.attachable` so that I could continue to enforce the one-to-one relationship between `Image` and `Source`, or between `Image` and `SourceSet` - for now.  Using `attachable` in this way seems to be an accepted Rails pattern for dealing with this type of polymorphic relationship.  As per our earlier discussion, we may reconsider the one-to-one relationship between `Image` and `attachable`, along with the one-to-one relationships between `Source` and other `asset`s.  I'll discuss with FA this afternoon.

I also fixed the line in `ApplicationController` where it was instantiating an object based on a string parameter received in the querystring.  Good catch!  The the [new method](https://github.com/dpla/primary-source-sets/commit/e2641df6accd85673cb9a4b66aafdb2b457c9dbe#diff-a7ae6992fe889ec0c720925f4275f636R67) specifies which params to use.

markbreedlove:
I think this is good to go -- ready to squash. We've discussed pretty well what we still have to work on, and I think it will be good to have this merged now so there's a multimedia foundation and iterate over various additions later.

AudreyAltman:
:squash:ed and rebased to develop

PR #25, opened 2015-09-29
add rake task to create sample admin account

AudreyAltman:
This introduces a rake task that creates a sample admin account for use in testing and development.


markbreedlove:
Thanks!  Looks like it works.

PR #26, opened 2015-09-29
edit data entry fields and forms

AudreyAltman:
This introduces some small edits to data entry fields and forms based on feedback from FA.  It addresses [#8055](https://issues.dp.la/issues/8055).

Field change:
* Increase the text limit for `guide.textual_content` from 65535 (MySQL TEXT) to 16777215 (MySQL MEDIUMTEXT)

Form changes:
* Change form label for `source_set.description` to "SEO description".
* Change form label for `source.name` to "Caption".
* Change form label for `source.textual_content` to "Description / transcription".

This also deletes a table definition from `db/schema.rb` that was mistakenly introduced in a previous PR.

### DIFF
--- a/app/views/source_sets/_form.html.erb
+++ b/app/views/source_sets/_form.html.erb
@@ -21,7 +21,7 @@
   </p>
  
   <p>
-    <strong><%= f.label :description %></strong><br/>
+    <strong><%= f.label 'SEO description' %></strong><br/>
     <em>Example: "This collection uses primary sources to explore the French and Indian War."</em><br/>
     <%= f.text_area :description %>
   </p>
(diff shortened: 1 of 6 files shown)


markbreedlove:
That looks good.

I have not thought about what database we should be targeting, though it probably doesn't matter that much for this project.  PostgreSQL might have different dimensions for different large-object datatypes.

markbreedlove:
This branch appears that it needs to be rebased before I can merge it.

AudreyAltman:
PostgreSQL just has one TEXT type of unlimited variable length, but if you don't specify a length in Rails it will default to an upper limit of 65535.  So I went with MySQL data types b/c, well, why not?

Rebase coming up.

AudreyAltman:
@markbreedlove - rebased and ready to :merge:

PR #27, opened 2015-10-01
Allow media file uploads to S3 buckets

markbreedlove:
Add file upload forms to the audio and video pages that post files directly to Amazon S3.

This is a first in a series of PRs that will add multimedia functionality.  After this come changes to enqueue transcoding jobs and to handle maintenance, like deletions and _maybe_ replacing files.

The instructions about setting up the S3 buckets are in-progress because there will be more things to set up related to transcoding and serving the files on the CloudFront CDN.

I don't know how to make tests for the Javascript code, but I mean to check out Jasmine and Teaspoon, which I had recommended to me recently.  I'm also not sure what spec tests I could have added that would have been helpful because the real action is in the Javascript and the s3_browser_uploads gem, which has its own tests.  I disabled some tests that require a rethink, because not all `create` resources in the app should issue HTTP redirects; RESTful ones should return "201" statuses.


app/assets/javascripts/application.js
```diff
@@ -14,3 +14,4 @@
 //= require jquery_ujs
 //= require turbolinks
 //= require_tree .
+//= require upload
```
AudreyAltman: I think that `require_tree .` will grab `upload.js`.  Is there a reason you're calling out `require upload` specifically?

markbreedlove: No, if that works, that's great. I just didn't know. I'll remove that line ...

app/views/audios/new.html.erb
```diff
@@ -1,5 +1,10 @@
+<script>
+    var pss_create_asset_path = '<%= source_audios_path() %>';
+    var pss_asset_type = 'audio';
+</script>
```
AudreyAltman: It might be cleaner to put this javascript in the `<head>` rather than the `<body>` (see http://stackoverflow.com/a/3438707)

markbreedlove: Thanks! I've changed it.

spec/support/shared_examples/nested_controller.rb
```diff
@@ -42,7 +42,8 @@
         end.to change(resource.class, :count).by(1)
       end
 
-      it 'redirects to the new resource' do
+      # FIXME:  this shouldn't be true, in the case of REST resources
+      xit 'redirects to the new resource' do
```
AudreyAltman: Instead of changing this spec, I would suggest opting out of the `:create` example for `audio` and `video`.   I tried to make this flexible on purpose so that we could customize as needed - you can specify which controller examples you want to use:  https://github.com/dpla/primary-source-sets/blob/develop/spec/controllers/audios_controller_spec.rb#L16 and https://github.com/dpla/primary-source-sets/blob/develop/spec/controllers/videos_controller_spec.rb#L16.  If you take out the `:create` param on these lines, you can just define whatever tests you want for the `create` action in `audio_controller_spec` and `video_controller_spec`.  Let me know if this makes sense.

I wrote a couple of `shared_examples` where you can opt in or out by controller action.  It was my attempt to streamline the process of writing controller specs by abstracting out common test patterns - it may or may not turn out to be helpful in the long run.

AudreyAltman:
@markbreedlove and I discussed that there seems to be a problem with the JavaScript event listener  where sometimes it doesn't correctly intercept the submit form action.  But otherwise this looks good.  I successfully uploaded an audio file and saw it saved to the database.

markbreedlove:
Removing `turbolinks` fixes the issue.  It appears to interfere with our events.

I've also made `upload.js` only be included on the pages that need it.  It was being included via `application.js` on all pages, which we don't want. I've removed the `require_tree .` from `application.js`. If we want to include files recursively, we can do so by putting them in specific directories, e.g. `common`.

markbreedlove:
https://github.com/dpla/primary-source-sets/commit/b6b677bbfb0f539fab79f1d570da5aaba5043824 should fix that problem with the disabled spec tests.

AudreyAltman:
I think this is ready to :merge: @markbreedlove do you want to :squash: first?

markbreedlove:
Yes, I will ...

PR #28, opened 2015-10-06
update schema.rb

AudreyAltman:
This updates `schema.rb` so its in line with the current migrations.


markbreedlove:
Discussed this w. Audrey and verified that the added columns, removed table, etc., are correct.

PR #29, opened 2015-10-06
manage assets through attachments table

AudreyAltman:
This introduces an `attachments` table to manage assets, and adjusts the relationships between `assets`, `sources`, and `sets` (see the updated [ER-Diagram](https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Relational+Database+Model) for reference).

What's changed?

* Before, there was a one-to-one relationship between `sources` and `assets` (ie. `images`, `videos`, `audios`, and `documents`).  Now, there is a many-to-many relationship between `sources` and `assets`.  This supports several use-cases we have recently realized we need to design for.

* Before, `source_sets` had a direct relationship with `images` (expressed through `images.attachable`).  Now, `source_sets` are related to `images` through `sources`. This change simplifies the database, and helps ensure that things like citation information (a property of `source`) will always be correctly associated with their `images`.

* `sources` has a new `featured` property.  In order for a `source_set` to get its "featured image" (ie. the image that is used in the UI to represent the set), it finds its associated `source` where `featured=true`, and then gets that `source`'s associated `image` where `size=small`.  

* Before, there were stricter validations on the models.  For example, `images` had a validation that ensured that each `source` had only one associated `image` of any given size.  Validations like these are more difficult now, so there is more impetus on the programmer and admin to keep track of all the associations.  We may be able to tighten up some of these restrictions after launch.

* `assets` are no longer nested under `sources` - nesting no longer makes sense with the many-to-many relationship between `assets` and `sources`.  Many of the changes to the `controllers` and `views` have to do with un-nesting the routes.

* `view` forms have been altered to support creating many-to-many relationships between `sources` and `assets`.

* The `basic_controller` `shared_example` spec has been refactored similarly to how the `nested_controller` was refactored in PR #27.  The re-directing behavior on the `:create` action has been moved to `redirecting_controller`.

* Before, due to a copy-paste error, many `reflect_on_association` specs had an erroneous `should` statement appended to the end - these have been removed.


markbreedlove:
I haven't fully reviewed this yet, but it looks like it should be merged before my changes that add media transcoding. Has it been rebased on `develop` as of today?

markbreedlove:
Never mind, the answer is yes, it has been rebased.

AudreyAltman:
This doesn't have to be merged before the transcoding stuff.  I can re-rebase after the transcoding stuff has been merged.

markbreedlove:
Either way will work.  I thought that your changes to the audios and videos controllers and models are more minor than mine, so it'll be easier for me to resolve merge conflicts than for you to.  Most of your changes are one-line changes to those files, while I am making multi-line changes.  Either way...

db/schema.rb
```diff
@@ -31,8 +31,15 @@
   add_index "admins", ["email"], name: "index_admins_on_email", unique: true
   add_index "admins", ["reset_password_token"], name: "index_admins_on_reset_password_token", unique: true
 
+  create_table "attachments", force: true do |t|
+    t.integer "source_id"
+    t.integer "asset_id"
+    t.string  "asset_type"
+  end
+
+  add_index "attachments", ["source_id"], name: "index_attachments_on_source_id"
```
markbreedlove: Is there a way in Rails migrations to have a compound index on three columns? That would be ideal here.

markbreedlove: At least, I think the asset_id should be indexed.

markbreedlove:
I've just been playing around with it. It's nice the way it works now with uploading the media assets.  I can add multiple videos and then I get a choice of which one to add to a source.

The selection is now implemented as radio buttons, which lets me pick multiple ones. I think that this actually wants to be a select where you can only choose one.

When there are a lot of videos added down the line, it might be a big list to scroll through, but I think that can be finessed later.

markbreedlove:
Thanks, @AudreyAltman, this looks good.  My only questions are regarding the compound index and the radio buttons instead of a select box.

AudreyAltman:
@markbreedlove I addressed your comments with two additional commits.  

I agree that radio buttons would be good - however, it proved more difficult than I anticipated to change these checkboxes into radio button due to some railsy expectations in the form helpers.  So instead, I wrote a little javascript that restricts users to choosing only one "main asset" (ie. document, audio, video, or large image), only one thumbnail, and only one small image per source.  Perhaps not ideal, but I think it will do.

I also added some indices on attachments.

AudreyAltman:
rebased to develop

markbreedlove:
Sounds good for now regarding the buttons.

With the compound index, can you confirm that the one-column index on `attachments.asset_id` should exist?  I can't think of when it would be used, considering the asset ID is only useful in conjunction with the asset type, in a query that joins it to the `sources` table; in which case, the compound index would be used, instead.

AudreyAltman:
@markbreedlove I was wondering about that myself.  I'll delete it!

AudreyAltman:
Removed index on `asset_id`: https://github.com/dpla/primary-source-sets/blob/feature/attachments-table/db/migrate/20151008142719_delete_attachments_index_on_asset_id.rb

:squash:ed and ready to :merge:

markbreedlove:
OK, schema.rb looks good, too.

PR #30, opened 2015-10-07
Add audio and video transcoding

markbreedlove:
Add media transcoding using the Brightcove Zencoder service.

To test this, you need a Zencoder API key, which requires a free signup. You can use your account's "integration" key to ensure that only test (non-billable) transcodings are performed.  Check `settings.local` for new fields that have been added.

Known potential issues:
* I'm not sure if the view helpers for the audio and video tags are ideal, because you have to specify `raw` when you call them from views, because they return strings of HTML.  But not that this has to hold things up right now.
* This is tied in to Zencoder.  The file upload in an earlier commit was also tied in to Amazon S3.  At least the `meta` fields and the generically-labeled `transcoding_job` fields of the `audios` and `videos` tables might make it not too hard to migrate if we ever have to.

### DIFF
--- a/app/helpers/transcoding_status_helper.rb
+++ b/app/helpers/transcoding_status_helper.rb
@@ -1,0 +1,8 @@
+module TranscodingStatusHelper
+  def transcoding_status(model_instance)
+    meta = JSON.parse(model_instance.meta)
+    s = meta.fetch('job', {}).fetch('state', false)
+    "<div class=\"trans-status\">Transcoding status: " \
+      "#{s ? s : 'pending'}</div>\n"
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ gem 'rubocop', '~> 0.32.1', require: false
 # came out in December, 2013, and I'm cautious about a change after this long
 # a time being disruptive. --MB
 gem 's3_browser_uploads', '0.1.2'
+gem 'zencoder', '~>2.5'
 
 group :test, :development do
   gem 'rspec-core', '~> 3.3.2'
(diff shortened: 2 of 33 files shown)


markbreedlove:
More known issues:

I added a number of methods to `ApplicationController` that I realize now should be Concerns in the `concerns` directory. It would be best if that could be done as a later refactor.  Related to that, there would be tests to change.

There is a test I've labeled `FIXME`, which needs to be moved to a spec for ZencoderAuthentication concern, but I'm not sure how to go about that at the moment.

I think it would be safe to take care of the shortcomings above with one or more post-launch tickets, to stay on schedule.

app/helpers/transcoding_status_helper.rb
```diff
@@ -0,0 +1,8 @@
+module TranscodingStatusHelper
+  def transcoding_status(model_instance)
+    meta = JSON.parse(model_instance.meta)
+    s = meta.fetch('job', {}).fetch('state', false)
+    "<div class=\"trans-status\">Transcoding status: " \
+      "#{s ? s : 'pending'}</div>\n"
+  end
+end
```
AudreyAltman: On my local instance, I am getting the message "Transcoding status: pending" even when the transcoding has completed.  I will dig in and see if I can find a cause.

markbreedlove: Did you run `zencoder_fetcher` against the correct `*_notifications` endpoint?

AudreyAltman:
I think this is ready to :squash:, rebase, and :merge:

markbreedlove:
OK, I'll do that, if you're sure you don't want to merge https://github.com/dpla/primary-source-sets/pull/29 first.

PR #31, opened 2015-10-07
Enable test coverage with simplecov

markbreedlove:
Add coverage reports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /config/database.yml
 Gemfile.lock
 .vagrant
+/coverage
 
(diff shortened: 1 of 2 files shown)


AudreyAltman:
Don't we also need to add `simplecov` to the `Gemfile`?

markbreedlove:
It's already in my Gemfile.lock, probably by virtue of being required by something else.

markbreedlove:
Apparently by `codeclimate-test-reporter`

AudreyAltman:
Ah!  In that case, I think this is good to go.

PR #32, opened 2015-10-08
add frontend_path method

AudreyAltman:
This adds a `frontend_path`, that constructs a link to the frontend.  It checks for the correct `/` between the path base defined in `Settings.frontend.url` and the appended path name.  This is more fool-proof than the previous code, which was unforgiving if both the path base and the appended path name did not have `/`s in the right places.  

It also moves the method out of the `source` model, to a more appropriate location in a `helper`.

This should fix the current problem on staging, where the path is showing up without the adjoining `/` (see [#8062](https://issues.dp.la/issues/8062))

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -23,6 +23,14 @@ module ApplicationHelper
     source.name.present? ? source.name : source.aggregation
   end
 
+  ##
+  # Returns frontend path with correctly joining '/'
+  # @param path String (with or without leading slash)
+  # @return String
+  def frontend_path(path = '')
+    Settings.frontend.url.chomp('/') + '/' + path.sub(/^\/+/) { $1 }
+  end
+
   private
 
   ##
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -2,20 +2,20 @@ require 'rails_helper'
 
 describe ApplicationHelper, type: :helper do
    
-   describe '#markdown' do
+  describe '#markdown' do
 
-     it 'renders HTML from a markdown String' do
-       text = '**Moomin!**'
-       expect(helper.markdown(text)).to eq("<p><strong>Moomin!</strong></p>\n")
+    it 'renders HTML from a markdown String' do
+      text = '**Moomin!**'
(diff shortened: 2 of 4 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -2,20 +2,20 @@
...
-       expect{ helper.markdown(invalid_input) }.to raise_error(TypeError)
+    it 'raises a TypeError if input is not a String' do
+      invalid_input = 123
+      expect{ helper.markdown(invalid_input) }.to raise_error(TypeError)
      end
-   end
+  end
 
-   describe '#inline_markdown' do
+  describe '#inline_markdown' do
```
AudreyAltman: I didn't change any of the tests in lines 1-18 (above), I just fixed the indentation.

app/helpers/application_helper.rb
```diff
@@ -23,6 +23,14 @@ def source_name(source)
     source.name.present? ? source.name : source.aggregation
   end
 
+  ##
+  # Returns frontend path with correctly joining '/'
+  # @param path String (with or without leading slash)
+  # @return String
+  def frontend_path(path = '')
+    Settings.frontend.url.chomp('/') + '/' + path.reverse.chomp('/').reverse
```
markbreedlove: Not that efficiency is a big issue here, but it would usually be more efficient to do the following, and may express more directly what's happening:  `.sub(/^/+) { $1 }`, such as:
```
irb(main):016:0* "/abc".sub(/^\/+/) { $1 }
=> "abc"
irb(main):017:0> "abc".sub(/^\/+/) { $1 }
=> "abc"
irb(main):018:0> "//abc".sub(/^\/+/) { $1 }
=> "abc"
irb(main):019:0> "//abc/".sub(/^\/+/) { $1 }
=> "abc/"
irb(main):020:0> "//abc/def".sub(/^\/+/) { $1 }
=> "abc/def"
irb(main):021:0> "abc/def".sub(/^\/+/) { $1 }
=> "abc/def"
```

AudreyAltman: Ah, this is good!  I will make the change.

AudreyAltman:
Change has been made here: https://github.com/dpla/primary-source-sets/pull/32/files#diff-127a850b7d7b20173ad3fba88f8a25e3R31 and :squash:ed

PR #34, opened 2015-10-09
Add Zencoder named credentials

markbreedlove:
Add handling for Zencoder named credential sets.  These are named sets of Amazon IAM (access management) keys that can be used in order to have separate permissions on development, staging, and production buckets.


markbreedlove:
I think I'll need to rebase this before merge.

markbreedlove:
I think that the ci check is stalled because the last commit is a `[ci skip]`.  I've seen it do this before.

AudreyAltman:
I think this looks good.  Ready to rebase and then :merge:

PR #35, opened 2015-10-09
Add Zencoder named credentials to outputs

markbreedlove:
A Zencoder credential name has to be added to each element of the `outputs` property of the job request, in addition to the top-level object, where it exists now.


AudreyAltman:
:+1:

PR #36, opened 2015-10-09
add dpla-branded header and footer

AudreyAltman:
This adds the DPLA-branded header and footer to all views.

This adds two new gems, `dpla_frontend_assets`, which has some images, fonts, and stylesheets; and `navigasmic`, which generates menus.

The following files were copied directly from the `frontend` app and have not been altered:
* `app/assets/fonts/*`
* `app/assets/images/*`
* `app/assets/app/assets/javascripts/*`
* `app/assets/stylesheets/frontend/*`
* `app/assets/stylesheets/vendor/*`
It's not ideal to have so many duplicate files in `frontend` and `primary-source-sets` and we can come up with a better way of managing them after launch.

There is a particularity in Rails 4 that requires some of the CSS from the `frontend` to be over-written.  The particularity is the way in which asset paths are expressed.  In Rails 3 an asset path would be, for example, `'/assets/icomoon.eot'`, whereas in Rails 4 it would be `'icomoon.eot'`.  The files `app/assets/stylesheets/frontend_overrides.css` and `app/assets/stylesheets/font_overrides.css` override asset path expressions in CSS files from the `frontend`.  I thought it better to create new files rather than overriding the `frontend` CSS files in place, because then we won't be maintaining different versions of the same files.  

The best way to test the views is to see them in your browser.


AudreyAltman:
I will look into why the tests are failing...

markbreedlove:
The "upload" (submit) button has disappeared on the audio / video upload forms, e.g. `/primary-source-sets/audios/new`.  In addition, on those pages, the search box's magnifying-glass icon is rendered twice, with one of the renditions being positioned incorrectly, as shown in the attached screenshot.  This is on Google Chrome.

Other than that, everything looks good!

I second your decision to avoid having modified versions of the `dpla_frontend_assets` files and agree that this is what we need to do for now.

screen shot 2015-10-09 at 12 51 42 pm

markbreedlove:
I'm observing the same things as above in the latest Firefox.

AudreyAltman:
The reason the CI test are failing is b/c Travis does not have access to the `dpla_frontend_assets`repo.  Running `bundle install --without dpla_branding` on Travis does not help b/c Bundler still wants to access `dpla_frontend_assets`, apparently to get its gemspec (https://github.com/bundler/bundler/issues/1552#issuecomment-4014684).  I'm still working on a solution.

AudreyAltman:
The [latest commit]
(https://github.com/dpla/primary-source-sets/commit/35f1da6b2232ed7e09e89cd59bff38e0c9b301e0) fixes the form submit button.  Still working on getting the Travis tests to pass.

AudreyAltman:
@mb and I have triumphed over Travis!!!  Hail the conquering heros!!!

AudreyAltman:
:squash:ed

PR #38, opened 2015-10-09
Add Image and PDF uploads

markbreedlove:


app/models/image.rb
```diff
@@ -1,8 +1,7 @@
 class Image < ActiveRecord::Base
   has_many :attachments, as: :asset
   has_many :sources, through: :attachments
-  validates :file_base, presence: :true
-  validates :mime_type, presence: :true
+  validates :file_name, presence: :true
   validates :height, numericality: { only_integer: true }
   validates :width, numericality: { only_integer: true }
```
AudreyAltman: The `:height` and `:width` validations here will cause the upload to fail if their value is `nil`.  We can solve this by adding `allow_blank: true`, so they would read:

`validates :height, numericality: { only_integer: true }, allow_blank: true`
`validates :width, numericality: { only_integer: true }, allow_blank: true`

markbreedlove: OK, I'll fix that.  I think I have it setting an empty string, but it's still a good thing to do this.

AudreyAltman:
This looks great.  I successfully an image and a pdf locally.  I only have one small suggestion, and that is to delete two remaining links I found to the now deleted `image:edit` and `document:edit` views.  They are in `app/views/images/show.html.erb` lines 4 and 5, and `app/views/documents/show.html.erb` lines 4 and 5.  

Great job getting this code done in such a short period of time!

AudreyAltman:
Also, it looks like this needs to be rebased to develop.

markbreedlove:
Aha, thanks for catching that.  Making the changes now.  Thanks for your kind words!  I'll rebase it after making the changes.

markbreedlove:
@AudreyAltman: It's rebased and squashed

AudreyAltman:
:dancer:

PR #39, opened 2015-10-12
Fix Content-Type of uploaded images and documents

markbreedlove:
Fix upload forms so that the Content-Type field assigned.  Amazon S3 defaults to "application/octet-stream" if it is not, and this causes PDFs to download instead of being displayed in the browser.  It also causes other files to be uploaded as octet-stream, but it's less noticable than with PDFs.

Also improve the frontend upload event-handler code to check that a file has been chosen.

### DIFF
--- a/app/assets/javascripts/docupload.js
+++ b/app/assets/javascripts/docupload.js
@@ -14,6 +14,16 @@
 
             e.preventDefault();
 
+            files = $('input[name=file]').prop('files');
+            if (! files.length) {
+                alert('Please choose a file.');
+                return false;
+            }
+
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/docupload.js
```diff
@@ -14,6 +14,16 @@
 
             e.preventDefault();
 
+            files = $('input[name=file]').prop('files');
+            if (! files.length) {
+                alert('Please choose a file.');
+                return false;
+            }
```
AudreyAltman: Nice!

AudreyAltman:
I tested this locally and it all seems to be working well.

PR #40, opened 2015-10-14
add form class

AudreyAltman:
This corrects an error introduced in PR #36.  On pages with both the DPLA search box and the asset upload form, the javascript could not differentiate between the two forms, causing both the search and upload actions to fail.  This adds a class name to the asset upload forms, and uses the class name in the jQuery selector to differentiate it from the DPLA search box.

This addresses [issue #8076](https://issues.dp.la/issues/8076).

### DIFF
--- a/app/assets/javascripts/docupload.js
+++ b/app/assets/javascripts/docupload.js
@@ -10,7 +10,7 @@
 
     $(document).ready(function() {
 
-        $('form').on('submit', function(e) {
+        $('form.upload').on('submit', function(e) {
 
             e.preventDefault();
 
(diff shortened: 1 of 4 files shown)


markbreedlove:
I should have mentioned that I was working on this in my Javascript refactoring!  :-/  Sorry.  It might be easier if I merge it with my upcoming PR, because I'm really rearranging the JS.

AudreyAltman:
Great minds think alike!  Sure, we can merge with your upcoming PR.

markbreedlove:
I'm just going to close this.  It's handled in PR#47.

PR #41, opened 2015-10-14
Fix/admin ui

AudreyAltman:
This fixes three bugs with the admin UI interfaces.

1. The `sources#show` view was failing if there were no associated media assets because the controller was trying to call `:file_name` on `nil`.

2. Submit buttons on several forms were not rendering correctly because they did not have a class name (they were rendering like the submit button in the DPLA search box).

3. On the `sources#show` view, link to the DPLA items were missing the `/item` substring.

### DIFF
--- a/app/controllers/sources_controller.rb
+++ b/app/controllers/sources_controller.rb
@@ -12,8 +12,12 @@ class SourcesController < ApplicationController
   def show
     @source = Source.find(params[:id])
     ma = @source.main_asset
-    @file_base_or_name =
-      ma.respond_to?(:file_base) ? ma.file_base : ma.file_name
+    @file_base_or_name = nil
+    
+    if ma.present?
+      @file_base_or_name =
+        ma.respond_to?(:file_base) ? ma.file_base : ma.file_name
+    end
   end
 
   def new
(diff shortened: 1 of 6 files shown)


markbreedlove:
That all looks right.  Would you like to squash and rebase it?

markbreedlove:
I just realized that there's no test and I'm not sure what the `#show` view will do with `@file_base_or_name` being undefined.  Looks like there might be an error rendering it.

AudreyAltman:
@markbreedlove when I run it locally, I'm not seeing a problem rendering `source#show`.  Can you give me specifics if you are getting an error?

markbreedlove:
Sorry I wasn't clear.  I haven't actually run it, but was just looking at the view, thinking there might be an issue with the variable not getting assigned in the case where `ma.present?` is false in the controller.

AudreyAltman:
Technically its assigned to `nil`: https://github.com/dpla/primary-source-sets/pull/41/files#diff-149237644d7fa1162622c0628311224aR15

But to answer your question, it isn't a problem in the view to render `file_base_or_name` when its value is `nil`.  Each time `file_base_or_name` is called, it is done in a context where `nil` is an okay value.

markbreedlove:
OK, thanks!  Want to squash it?

AudreyAltman:
yes!

AudreyAltman:
:squash:ed

PR #42, opened 2015-10-14
order sources and sets by created_at

AudreyAltman:
On `sets#show` order `sources` by their created date (earliest created should appear first).
On `sets#index`, order `sets` by their created date (earliest created should appear first).
This addresses [ticket #8078](https://issues.dp.la/issues/8078)

### DIFF
--- a/app/views/source_sets/index.html.erb
+++ b/app/views/source_sets/index.html.erb
@@ -3,7 +3,7 @@
 <p><%= link_to "Create new set", new_source_set_path %></p>
 
 <table>
-<% @source_sets.each do |set| %>
+<% @source_sets.order('created_at ASC').each do |set| %>
   <tr>
     <td><%= inline_markdown(set.name) %></td>
     <td><%= link_to "View", source_set_path(set) %></td>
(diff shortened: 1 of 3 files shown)


AudreyAltman:
I will investigate this failing test as soon as I can.

AudreyAltman:
Test is passing.  Ready for review.

markbreedlove:
Looks simple enough.  I'm noticing the change to `index.html.erb_spec.rb`. Does that test check whether they are ordered?

AudreyAltman:
It does not check that they are ordered, it just changes the value of `@source_sets` in the spec from an `Array` to an `ActiveRecord::Relation` (in the real app, the value of `@source_sets` will be an `ActiveRecord::Relation`).  The `Array` was doing a good job of emulating the true value of `@source_sets` before b/c they are both enumerables.  But, `Array` does not respond to the `order` method, so it was causing the test to fail.

AudreyAltman:
@markbreedlove do you think I should add a test that checks for order?

markbreedlove:
No, not now. I'll go ahead and merge it if it's rebased.

PR #44, opened 2015-10-14
fix featured image attr name

AudreyAltman:
This fixes the attribute name of the featured image on the `source_sets#show` view.  The error was preventing the page from rendering.

This addresses [ticket #8079](https://issues.dp.la/issues/8079).

### DIFF
--- a/app/views/source_sets/show.html.erb
+++ b/app/views/source_sets/show.html.erb
@@ -38,7 +38,7 @@
     <td><strong>Featured image </strong></td>
     <td>
       <% if @source_set.featured_image.present? %>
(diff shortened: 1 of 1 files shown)


markbreedlove:
Looks good.  I'll merge if you rebase.  I'm tempted to use the "Update branch" button on this page, but didn't you have some issue with that before?

AudreyAltman:
I have used the "Update branch" button.  I don't remember exactly what it does, but @no-reply and I decided that doing the rebase locally is better.

AudreyAltman:
@markbreedlove rebased and ready to :merge:

PR #48, opened 2015-10-14
Fix Source has_many relationship with Attachment

markbreedlove:
Have a deletion of a Source record cause a deletion of the corresponding Attachment record.


AudreyAltman:
:+1:

PR #49, opened 2015-10-15
Add Google Analytics

markbreedlove:
Add Google Analytics tracking code.

It's enabled by setting `googleanalytics.tracker` in `settings.local.yml` to a string value.

I've only added it to the views for `/sources/:id` and `/sets/:id` because I'm not sure which ones are public, versus administrative.

@AudreyAltman, can you tell me what pages should be tracked, or at least help me identify which ones are eventually going to be public?  I'm not clear on this because of the presence of certain links that look like they should only be on admin pages.  It does appear that `/` (`source_sets#index`) would be another candidate.

### DIFF
--- a/app/views/guides/show.html.erb
+++ b/app/views/guides/show.html.erb
@@ -1,3 +1,7 @@
+<% content_for :foot_script do %>
+<%= render partial: 'shared/analytics' %>
+<% end %>
+
 <h1>Teaching guide: <%= inline_markdown(@guide.name) %></h1>
 
 <p>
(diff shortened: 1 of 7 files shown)


markbreedlove:
@AudreyAltman: that last commit adds analytics to more views.

The total list of pages where users will be tracked is:
* `/` or `/sets` (`source_sets#index)
* `/sets/:id` (`source_sets#show`)
* `/sources/:id` (`sources#show`)
* `/guides/:id` (`guides#show`)

I'll squash and rebase if that is going to be OK.

AudreyAltman:
Once its squashed and rebased, this looks ready to go.

PR #50, opened 2015-10-15
Feature/source style

AudreyAltman:
This styles the `source#show` view for video, audio, and document assets.  The best way to test will be to view them in your browser window.

In the process of adding the PDF viewer, I found that the `source` views needed access to what was the `@base_src` var in the `audio` and `video` controllers - so I changed it to an `ApplicationHelper` method and edited the specs accordingly.

It includes one JavaScript function for dynamic styling. I couldn't find a way to make this particular style definition work with static CSS.  It's a minor style enhancement that doesn't affect the actual content, so if a browser has JavaScript disabled, the user will get a comparable experience.

The video player now has a responsive height and width, based on the height and width of the browser window, instead of being set at 640px by 480px.  If we find that video quality is degrading when the video player is too large, we can add a `max-height` and `max-width` to the CSS.

Rendering image assets with OpenSeadragon can be done later in `views/sources/_image.html.erb`.

Known issue: The PDF viewer doesn't render if JavaScript is disabled.  We can add a graceful degradation that provides a link to the PDF instead.

### DIFF
--- a/app/assets/javascripts/style.js
+++ b/app/assets/javascripts/style.js
@@ -1,0 +1,25 @@
+/*
+ * Handles dynamic styling of the DOM.
+ */
+(function() {
+
+  $(document).ready(function() {
+    setAsideHeight();
+  });
+
+  $(window).resize(function() {
+    setAsideHeight();
+  })
+})();
+
+/*
+ * Extend the height of the aside module to be at least as tall as the rendered
+ * media asset.
+ */
+function setAsideHeight() {
+  var height = $('.source aside .module').outerHeight();
+  var minHeight = $('.source .media-inner-container').height();
+  if (minHeight > height) {
+    $('.source aside .module').outerHeight(minHeight);
+  }
+}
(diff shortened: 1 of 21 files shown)


markbreedlove:
Thanks for `base_src`!

I think the `max-width` and `max-height` properties would be a good idea, considering some of the content I've seen being added.

The `<aside>` is a little on the tall side now in my browser but I'm not sure how important that is to change considering the time constraints now.

markbreedlove:
I'd suggest 1280 x 720 as the maximum dimensions of the video window.

If users want to see it full-screen, they can click on the full-screen button.  This full-screen feature (at least in Chrome) is going to be pulled some day if it's not an https resource, but this should be OK for now.

AudreyAltman:
@markbreedlove : I added `max-height` and `max-width` to video: https://github.com/dpla/primary-source-sets/pull/50/files#diff-1af037e133c48e79b226e4a29de17643R53

:squash:ed and rebased to develop.

PR #51, opened 2015-10-16
add destroy depenedency to assets-attachments reltionship

AudreyAltman:
This adds a destroy dependency to the assets-attachments relationship so that when an asset is deleted, its rows in the attachments table are also deleted.


markbreedlove:
That looks good. I'll merge it if you can confirm that you've observed how it works in the database console.

AudreyAltman:
Yes, I have experimented with it in the console, and it all works as expected.

markbreedlove:
OK!

PR #53, opened 2015-10-16
style source_sets#show

AudreyAltman:
This introduces CSS styling for `source_sets#show`.  It includes one helper method for concatenating multiple authors and their affiliations.  It also includes two new methods in the `Source` model to ease access to image assets of specified sizes.  Finally, it corrects a small HTML error in `sources#show`.

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -2,6 +2,85 @@
   word-wrap: break-word;
 }
 
+.primary-source-sets aside {
+  margin-left: 1em;
+  margin-bottom: 1em;
+}
+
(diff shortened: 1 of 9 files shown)


markbreedlove:
That all looks good.  Is this the place for questions about design details, like the alignment of boxes on the page, or should that wait until later?

AudreyAltman:
@markbreedlove I would love your feedback about design details, but unless you think they are detrimental to user experience, I am thinking that they could be made into new tickets that I can try to get to on Monday.  It would be good to get something rough out the door today and finess it later.

markbreedlove:
Great, that's what I thought.  This is good to go, then.

PR #55, opened 2015-10-16
style teaching guide

AudreyAltman:
This introduces CSS styling for a guide.

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -1,3 +1,5 @@
+/* Shared styles */
+
 .wrap {
   word-wrap: break-word;
 }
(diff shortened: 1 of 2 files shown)


markbreedlove:
If there could be a style for `.primary-source-sets .moduleSection a` after [this point](https://github.com/dpla/primary-source-sets/blob/feature/guide-style/app/assets/stylesheets/primary_source_sets.css#L12-L14), then it would keep the anchor links from wrapping:
```
white-spece: nowrap;
```
If you want to take a minute to do that.  If not, this is fine; happy to merge it.

AudreyAltman:
@markbreedlove which anchor links are you thinking that we should keep from wrapping?  Some of the links in the modules are longer than the width of a line, so I think we need to wrap them.

AudreyAltman:
I am going to change the hard-coded email address in this view as well.

markbreedlove:
In that case, this can be revisited later.

markbreedlove:
I was thinking the link in the box that says "Primary Source Set" in the upper-right.

AudreyAltman:
Oh yeah, some of those Primary Source Set names are super long.

AudreyAltman:
Okay, I changed the previously hard-coded [email](https://github.com/dpla/primary-source-sets/commit/8078eea05f53f38ae689679d6c07fa8723d35e00#diff-d18bc7e6b81d90f582b475cfa9d0c18aR90).  I also added some [conditional statements](https://github.com/dpla/primary-source-sets/commit/8078eea05f53f38ae689679d6c07fa8723d35e00#diff-d18bc7e6b81d90f582b475cfa9d0c18aR30) so that "questions" and "activity" only display if they are present.  I tested both changes in my browser.

markbreedlove:
:+1:

PR #57, opened 2015-10-16
style source_sets#index

AudreyAltman:
This styles `source_sets#index`.

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -40,6 +40,28 @@
   font-weight: bold;
 }
 
+/* SourceSets index styles */
+
+.all-sets .set-list {
+  padding-bottom: 2em;
+}
+
+.all-sets .module {
+  border-color: #6592A6;
+  padding: 0;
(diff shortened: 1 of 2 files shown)


AudreyAltman:
I should note that all the `featured_image`s (ie "small" images) are the same height and width dimensions, and they are slightly larger than the largest possible zoom in this layout, so it should look good.

AudreyAltman:
rebased to develop

AudreyAltman:
A caveat to my earlier comment about the images looking good in the layout - they are a off-center in the mobile view, but I think this is something we can fix next week.

markbreedlove:
I'm forget how the thumbnails or small images are supposed to bubble up to the set list, but it's probably OK if you've tested it.  I have an index page without any thumbnails, but I can't find where to assign a thumbnail for the set itself.

AudreyAltman:
You have to assign a "small" image to a source, and then mark that source as "featured".  Then the set that is the parent of that source will use the "small" image on this page. But yes, I've tested it.

markbreedlove:
ok

PR #58, opened 2015-10-16
Add breadcrumbs

markbreedlove:
Add breadcrumb navigation to public-facing pages:  source sets, sources, and guides.

Do we need a link to the home page of the main site?

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -4,6 +4,7 @@
 # @see Guide
 class GuidesController < ApplicationController
   before_filter :load_source_set, only: [:index, :new, :create]
+  add_breadcrumb 'Primary Source Sets', :root_path
 
   def index
     redirect_to @source_set
@@ -11,6 +12,8 @@ class GuidesController < ApplicationController
 
   def show
     @guide = Guide.friendly.find(params[:id])
+    add_breadcrumb @guide.source_set.name, source_set_path(@guide.source_set)
+    add_breadcrumb 'Guide', guide_path(@guide)
   end
 
   def new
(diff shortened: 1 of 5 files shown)


markbreedlove:
(The "info" section of the site does _not_ have a breadcrumb link to the home page.)

markbreedlove:
I'm inclined to think "no" because there are already multiple links to the home page elsewhere in other navigation elements.

app/controllers/sources_controller.rb
```diff
@@ -13,6 +14,8 @@ def index
 
   def show
     @source = Source.find(params[:id])
+    add_breadcrumb @source.source_set.name, source_set_path(@source.source_set)
+    add_breadcrumb @source.name, source_path(@source)
```
AudreyAltman: Instead of the `@source.name` here, could we just have the text of the link be "Source"?  Some of the source names are several hundred characters long, and that's a lot to have in a breadcrumb.

markbreedlove: Sure, that's fine.  Changing it ...

AudreyAltman:
I agree that we should not have a link to the homepage in the breadcrumbs.

app/controllers/guides_controller.rb
```diff
@@ -11,6 +12,8 @@ def index
 
   def show
     @guide = Guide.friendly.find(params[:id])
+    add_breadcrumb @guide.source_set.name, source_set_path(@guide.source_set)
+    add_breadcrumb @guide.name, guide_path(@guide)
```
AudreyAltman: Like my source comment (below), could we have the text of the link be "Teaching guide" instead of `@guide.name`?  Again, because of the length of the names.

markbreedlove: OK, changing ...

AudreyAltman:
Aside from my two suggestions above, I think this looks great!

markbreedlove:
OK, if that looks good, then I'll just squash this in a minute.

AudreyAltman:
:+1:

PR #59, opened 2015-10-19
Disable context menu on lightbox links to big images

markbreedlove:
This gets rid of the context menu (right click menu) of Lightbox's links to big images.  We had previously removed the ability to right-click and "download image as" but this gets rid of the menu with the options to open the link in another window or "save link as."

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,3 +12,10 @@
 //
 //= require jquery
 //= require jquery_ujs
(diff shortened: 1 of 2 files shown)


AudreyAltman:
This looks good.  I tested it on my browser and it worked as expected.  One thing to note is that on browsers with javascript disabled, this fix won't work - but I think that for the sake of getting to launch, this is okay.

PR #63, opened 2015-10-20
Fix sign-in page server error

markbreedlove:
Move breadcrumb link rendering to just those three views that need it, taking it out of the application layout.

The presence of a `url_for` call that asked for the location of `source_sets#index` caused a server error in the context of a Devise login page, with an error indicating that no route matches `{:action=>"index", :controller=>"devise/source_sets"}`.


AudreyAltman:
:+1:

PR #65, opened 2015-10-20
UI presentational fixes for launch

AudreyAltman:
This makes some presentational adjustments to the public UI for launch.  Changes include:

`source_sets#index`
* Remove paragraph break in intro text
* Remove period at end of byline
* Fix width of set tiles for mobile view
* Make bottom margin of first aside module smaller

`source_sets#show`
* Eliminate text wrapping in intro text

`guides#show`
* Make title bar and first `aside` module the same height (this is done dynamically with JavaScript, which is not ideal but will get us to launch)
* Remove period at end of byline
* Make bottom margin of first aside module smaller

### DIFF
--- a/app/assets/javascripts/style.js
+++ b/app/assets/javascripts/style.js
@@ -23,13 +23,25 @@ function setAsideHeight() {
     $('.source aside .module').outerHeight(minHeight);
   }
 
-  var moduleHeight = $('.set aside .module').outerHeight();
+  // resize header on set page
+  var moduleHeight = $('.set .guide-link .module').outerHeight();
   var titleHeight = $('.set .title-outer-container').outerHeight();
 
   if (moduleHeight < titleHeight) {
-    $('.set aside .module').outerHeight(titleHeight);
+    $('.set .guide-link .module').outerHeight(titleHeight);
   }
   if (titleHeight < moduleHeight) {
     $('.set .title-outer-container').outerHeight(moduleHeight);
+  }
+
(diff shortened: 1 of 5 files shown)


AudreyAltman:
rebased to develop

markbreedlove:
If I load `guides#show`, then resize the page smaller, the gray background of the "Primary Source analyisis" div becomes improperly sized (see screenshot).  If I reload the page, it's fine.  If I resize the window larger, it's fine.
screen shot 2015-10-20 at 12 44 15 pm

AudreyAltman:
@markbreedlove yeah, I'm aware that's happening and its less than ideal - but maybe something we can address after launch?  Unless you know how to fix it...

markbreedlove:
That's in Chrome.  In firefox it exhibits the same behavior if I make the window smaller _or_ bigger.

markbreedlove:
OK, yes, if you know about it already, then let's get it after launch.

AudreyAltman:
I think it's b/c of the javascript I'm using to make the blue title bar and the "Primary Source Set" module the same size.  Hang on, I might know how to fix it...

markbreedlove:
The other pages look fine, given the degree to which I can test them right now.  (E.g. have not looked on a mobile device.)

markbreedlove:
OK, holding ...

markbreedlove:
It's fixed.  Will merge when tests pass.

PR #71, opened 2015-11-02
title, description, & canonical link tags

AudreyAltman:
This adds unique SEO tags to the `head` of public-facing views.

1. A `title` tag.
2. A `meta` `description` tag.
3. A `canonical` `link` tag, which should be an absolute URL with a trailing slash.  `source_set#index` does not have a `canonical` `link` tag because in future iterations, the contents of this page will change based on search parameters (ie. there will not be only one version of the page).

The contents of these tags are text-only, so they are stripped of any nested HTML tags where appropriate.


AudreyAltman:
Rebased to develop

PR #73, opened 2015-11-10
fix title style and content spacing

AudreyAltman:
This makes some small changes to the style of the `guide`.

It removes the blue background from the title and byline, instead making the title blue text on a white background.  It also adds more whitespace in-between sections of the main content of the page.

This addresses [ticket #8150](https://issues.dp.la/issues/8150).

### DIFF
--- a/app/assets/javascripts/style.js
+++ b/app/assets/javascripts/style.js
@@ -33,15 +33,4 @@ function setAsideHeight() {
   if (titleHeight < moduleHeight) {
     $('.set .title-outer-container').outerHeight(moduleHeight);
   }
-
-  // resize header on guide page
-  var moduleHeight = $('.guide .set-link .module').outerHeight();
-  var titleHeight = $('.guide .title-outer-container').outerHeight();
-
-  if (moduleHeight < titleHeight) {
-    $('.guide .set-link .module').outerHeight(titleHeight);
-  }
-  if (titleHeight < moduleHeight) {
-    $('.guide .title-outer-container').outerHeight(moduleHeight);
-  }
 }
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -20,23 +20,6 @@
 
 /* SourceSet and Guide shared styles */
 
-.set .title-outer-container,
-.guide .title-outer-container {
-  background-color: #6592A6;
-  margin-bottom: 2em;
-}
-
-.set .title-inner-container,
-.guide .title-inner-container {
-  color: white;
-  padding: 1.5em;
-}
-
-.set h1, .guide h1 {
-  color: white;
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/style.js
```diff
@@ -33,15 +33,4 @@ function setAsideHeight() {
...
-  // resize header on guide page
-  var moduleHeight = $('.guide .set-link .module').outerHeight();
-  var titleHeight = $('.guide .title-outer-container').outerHeight();
-
-  if (moduleHeight < titleHeight) {
-    $('.guide .set-link .module').outerHeight(titleHeight);
-  }
-  if (titleHeight < moduleHeight) {
-    $('.guide .title-outer-container').outerHeight(moduleHeight);
-  }
```
AudreyAltman: This javascript resized the blue background box for the title.  Since the title no longer has a blue background, the javascript is no longer needed.

app/assets/stylesheets/primary_source_sets.css
```diff
@@ -55,6 +38,21 @@
...
+
+.set .title-inner-container {
+  color: white;
+  padding: 1.5em;
+}
+
+.set h1 {
+  color: white;
+  margin: 0;
+}
```
AudreyAltman: I moved this CSS so that it would be under the `/* SourceSets index styles */` header instead of the `/* SourceSet and Guide shared styles */`.

app/views/guides/show.html.erb
```diff
@@ -28,19 +28,23 @@
       </div>
     </div>
 
-    <div class='textual_content'>
+    <div class='textual-content'>
```
AudreyAltman: This just corrects a typo - all the other class names use `-` instead of `_`.

markbreedlove:
For some reason I'm getting the `h1` title with black text instead of blue.  The color is `#222` as defined by https://github.com/dpla/primary-source-sets/blob/feature/tweak-teaching-guide-style/app/assets/stylesheets/vendor/main.css#L245

AudreyAltman:
Is your local deployment getting the assets from the `frontend-assets` gem?  That might be the reason why the `h1` is black instead of blue.  If you look at `/primary-source-sets`, is the `h1` on that page also black?  If so, that could be an indication that the `frontend-assets` aren't loading.

markbreedlove:
No, that must be the issue.  I have it running without frontend-assets.  Is that OK, that it depends on that, and shows black instead of blue just because the assets aren't loaded?

AudreyAltman:
I think it has to be okay - a lot of the CSS is defined in the `frontend-assets`, so unless we want to duplicate the stylesheets for all the frontend apps, I think it's unavoidable.

markbreedlove:
OK, looks good with the frontend-assets gem in-place.

PR #74, opened 2015-12-01
Fix/json ld examples

AudreyAltman:
Update the JSON-LD examples to reflect the changes made in the lead-up to launch, and to correct errors in schema.org compliance.  Also, consolidate the three separate `.json` example files into one `.json` file for easier maintenance.

Changes include:
* Remove `SourceSet.image` because we decided not to support it.
* Correct typos related to the `authors` field in `TeachingGuide`.
* Change `Source.@type` from `MediaObject` to `CreativeWork` for compliance with schema.org definitions.
* Change `Source.encodingFormat` to `Source.fileFormat` for compliance with schema.org definitions.
* Use `Source.mainEntity` to describe the source object.
* Associate `MediaObject`s with the `mainEntity`.  This will allow us to have multiple digital representations of the same conceptual entity (for example, a video clip and a script of the same film).

You can view commit [bd0214e](https://github.com/dpla/primary-source-sets/commit/bd0214e55ea107b6fbc9f872da8ff3fc174dd334) to see easily what changed in the Source Set and Teaching Guide sections of the data model.

This addresses [ticket #8175](https://issues.dp.la/issues/8175).


anarchivist:
Looks good, but it appears there's a (consistent) typo throughout the file (`tagetUrl` instead of `targetUrl`).

AudreyAltman:
Typos fixed and :squash:ed.  Will rebase once PR#86 has been merged.

anarchivist:
@AudreyAltman :+1:

anarchivist:
Looks like we'll also need to change the `@context` to be something like this, assuming we want Schema.org to be the default vocabulary:

```json
{
   "@context": {
      "@vocab": "http://schema.org/",
      "ccss": "http://www.corestandards.org/" # ...
   }
}
```

AudreyAltman:
@anarchivist good catch!  I've fixed the `@context` and rebased to develop.

PR #75, opened 2015-12-02
do not return nil assets in Source#assets

AudreyAltman:
This ensures that `Source#assets` does not return any `nil` values.  `nil` values may occur if someone deletes an asset (an `Image`, `Document`, `Video`, or `Audio`) without deleting its association to a `Source`, as expressed in the `attachments` table.  There are validations in place to dependent-destroy these associations, but this in an extra precaution that will catch any mistakes that have occurred or could occur in the future.

This addresses [task #8167](https://issues.dp.la/issues/8167).

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -39,7 +39,7 @@ class Source < ActiveRecord::Base
   # Gets all assets associated with source through attachments table.
   # @return Array<ActiveRecord::Base>
   def assets
-    attachments.map(&:asset)
+    attachments.map(&:asset).reject{ |a| a.nil? }
   end
 
   ##
--- a/spec/factories/attachments.rb
+++ b/spec/factories/attachments.rb
@@ -1,0 +1,7 @@
+FactoryGirl.define do
+  factory :attachment, class: Attachment do
+    source_id nil
+    asset_id nil
+    asset_type nil
+  end
+end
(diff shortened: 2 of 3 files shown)


markbreedlove:
This looks fine, but can you remind me of what `&:asset` is in the context of [#assets](https://github.com/dpla/primary-source-sets/blob/61d345ea6fa01704988e089f75660f66b756038b/app/models/source.rb#L42)?

I'm not able to look at it and recall where the definition of this proc (?) is or what kind of values it returns.
Is this the normal way that Rails lets you work with many-to-many joins, or left joins?
I'm assuming there' some Rails "magic" that is not obvious to me.

AudreyAltman:
Sure.  The `attachments` table has three columns, `source_id`, `asset_id`, and `asset_type`.  The `asset_id` refers to a `Video`, `Image`, `Document`, or `Audio`, and  `asset_type` holds the class name of the asset.  So `&:asset` is going to return an instance of `Video`, `Image`, `Document`, or `Audio`.  If the value for `asset_id` doesn't match up with any records that are saved in the database, it will return `nil`.  Let me know if that doesn't answer your question.

markbreedlove:
I think I understand, and it seems that I'm taking this out of the scope of the PR, but I'm just wanted to make sure it's not something that needs documentation.  Is that Rails magic for a column that ends with `_id`?  My concern in looking at it is just that, though I can tell it's rejecting nil values I have no understanding of where this thing `asset` comes from or what values it's supposed to return -- but that's just my general gripe about Rails in general.

AudreyAltman:
@markbreedlove yes, it is Rails magic, based entirely on the naming conventions.

markbreedlove:
OK, thanks, @AudreyAltman.  I'll merge this.

PR #76, opened 2015-12-02
Allow admins to designate and view unpublished sets

AudreyAltman:
This allows admins to assert whether or not `source_sets` are "published".  Published sets are visible to all users from `source_sets#index`; unpublished sets are visible only to admins.  The `source_sets` table has a column, `published`, that takes a boolean value.  Up to this point, the default has been to set all `source_sets` as `published=true` in the database.

Changes include:
* A place on the `source_sets` form to specify whether a set is "published" or not.
* Model methods that get published and unpublished sets from the database.
* Setting instance variables in the `source_sets#index` controller for `@published_sets` and `@unpublished_sets`.
* Abstracting out a view partial that renders a list of sets, be they published or unpublished.
* Changing the `source_sets#index` view to show unpublished sets only to admins.

This PR purposely leaves all sets visible to all users when accessed directly with the URL.  This will allow members of the EAC to view unpublished sets before we build out the authentication system to create view-only access (see [task #8163](https://issues.dp.la/issues/8163)).

If you view `source_sets#index` in your browser, you will notice that the unpublished sets have a different background color than the sets.  This is intentional.

This addresses [task #8102](https://issues.dp.la/issues/8102).

### DIFF
--- a/app/controllers/source_sets_controller.rb
+++ b/app/controllers/source_sets_controller.rb
@@ -8,7 +8,8 @@ class SourceSetsController < ApplicationController
   before_action :authenticate_admin!, only: [:new, :edit]
 
   def index
-    @source_sets = SourceSet.all
+    @published_sets = SourceSet.published_sets
+    @unpublished_sets = SourceSet.unpublished_sets
   end
 
   def show
@@ -58,6 +59,7 @@ class SourceSetsController < ApplicationController
                                        :description,
                                        :overview,
                                        :resources,
+                                       :published,
                                        author_ids: [])
   end
 end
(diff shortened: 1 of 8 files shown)


AudreyAltman:
This is assigned to @markbreedlove for review, but it could also be reviewed by @no-reply or @anarchivist.  It's pretty basic Rails logic that doesn't require intimate knowledge of the app.

no-reply:
This looks rock solid to me. Nice, clean, well tested code. :+1:

AudreyAltman:
:dancer: thanks @no-reply

PR #78, opened 2015-12-07
add save button to top of set, source, and guide forms

AudreyAltman:
This adds "submit" buttons to the tops of data entry forms for `source_sets`, `guides`, and `sources`.  When tested locally, these buttons work as expected.

This addresses [ticket #8101](https://issues.dp.la/issues/8101).

### DIFF
--- a/app/views/guides/_form.html.erb
+++ b/app/views/guides/_form.html.erb
@@ -14,6 +14,8 @@
     </div>
   <% end %>
 
+  <%= f.submit 'Submit', class: 'form-submit' %>
+
   <p>
     <strong><%= f.label "Name (required)" %></strong><br/>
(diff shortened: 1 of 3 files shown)


markbreedlove:
Is the wrapping of buttons in `<p>` tags done for consistency with views elsewhere? Since a button is a block-level element, it shouldn't need it.

markbreedlove:
Otherwise, I'm happy to merge this if removing the `p` elements is going to cause trouble with the css. Maybe another ticket can come along later to clean that up.

AudreyAltman:
Good point about the `<p>` tags.  I'll remove them real quick.

AudreyAltman:
Removed `<p>` tags and rebased to develop.

PR #79, opened 2015-12-07
reduce whitespace below guide byline

AudreyAltman:
Reduce the whitespace on `guides`, below the byline and above the `.textual-content` block.  This addresses [note #6 in task #8150](https://issues.dp.la/issues/8150#note-6).  I tested it locally, and also showed it Franky for approval.

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -121,7 +121,11 @@
 
 /* Guide styles */
 
-.guide .textual-content, .guide .questions, .guide .activity {
+.guide .textual-content {
+  margin-top: 1.5em;
+}
+
+.guide .questions, .guide .activity {
   margin-top: 3em;
 }
 


markbreedlove:
If ems are used in general, this wouldn't be the time to change it, so I wouldn't want to make that suggestion, but I'm more inclined to use pixels for whitespace and layout, because the spacing that you need to make things look right doesn't scale directly in proportion to the default text size that is set up in the browser.  (The whitespace isn't directly related to the hight of a font; it has to do with a lot of other things and it doesn't go in proportion to just one factor.)  If anyone's changed the default font size, it's more likely that everything's going to shift into weird places with layouts specified in ems. I recall being motivated years ago by something I read on accessibility or device interoperability to use ems for everything, and I did for a while, but it never made my pages more useful to more people.

For the case at hand, this fine!  I'll merge it.

AudreyAltman:
@markbreedlove FWIW I tend to use ems b/c I found that pixels can be unpredictable on high vs. low-density mobile screens.  But I haven't actively tested that in a couple years so maybe things have changed.  Would be worth experimenting and considering what's best to use moving forward.

markbreedlove:
I always assume that, for the mobile screens, I want to use alternate stylesheets.  I can't remember the exact syntax for that and can't remember if you can detect pixel density but everything's going to need different layout decisions on a mobile device anyway, so I just figure it'll get handled with an alternate stylesheet.

PR #81, opened 2015-12-10
clean up global variables

AudreyAltman:
This cleans up the use of global and Fixture variables and in the jasmine tests.

In this app, some JavaScript variables are set in views (ie. `pss_asset_type`).  Before, these variables were simulated in Jasmine tests by adding them to Fixtures with HTML `<script>` tags.  For example:

    <script>var pss_asset_type = "video";</script>

While this accurately replicates the way in which these variables are set the app, the difficulty with this approach is that when Jasmine runs its `cleanUp` function between tests, the JavaScript variables are not cleaned up, so they remain set for _all_ subsequent tests.

This introduces a different approach.  The variables set in views are introduced as global variables for the Jasmine tests .  Since they are global variables, they can be set in Before blocks, accessed by individual tests, and deleted in After blocks (local variables cannot be deleted in JavaScript outside of garbage collection).

This PR also scopes variables that do not need to be global to keep the namespace clean and minimize potential confusion.

### DIFF
--- a/spec/javascripts/av_upload_handler_spec.js
+++ b/spec/javascripts/av_upload_handler_spec.js
@@ -1,17 +1,22 @@
 
 describe("AVUploadHandler", function() {
+    var a;
 
     beforeEach(function() {
         a = new AVUploadHandler();
         spyOn(a, 'postAssetRecord').and.stub();
-        html = '<script>var pss_asset_type = "video";</script>' +
-               '<form></form>';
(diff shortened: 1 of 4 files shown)


markbreedlove:
OK, that looks good! I need to remember to use `var` consistently.

I will merge this after it's rebased.

AudreyAltman:
@markbreedlove rebased to develop

PR #82, opened 2015-12-10
Allow creation of media assets with specified source.

AudreyAltman:
This allows creation of media assets (`image`, `document`, `video`, or `audio`) with a specified `source`. 

Sources and assets have a many-to-many relationship in the database.  Previously, admins could only persist relationships between assets and sources using the form on the `sources#new` view to select existing assets from a menu. This made workflow difficult for admins.

This PR introduces a way to specify which source an asset is associated with upon the asset's creation.  A source can be specified in a `source_id` parameter in the asset's `#new` route.  When the asset is saved, its association with the source is saved as well.

And important particularity of this approach is that _upon creation_, an asset can only be associated with one source (there are other ways to build additional associations after an asset is created).  However, the vast majority of assets are associated with only one source, so this approach should have a big positive impact on admin's workflow.

Changes include:

* In `sources#show`, add links to `asset#new` routes that include the source's id as a parameter.  For example, if you are a logged-in admin on `/sources/4`, when you click on "Add a new image", you will be taken to `/images/new?source_id=4`. 

* Allow the asset controllers to deal appropriately the `source_id` parameters.  For the `#new` method this means instantiating a `Source` object from the `source_id` param and making it accessible to the view.  For the `#create` method, this means permitting the `source_ids` param in a way that works with both the JavaScript and the asset models.

* On the `asset#new` view, set the `source_id` as a JavaScript variable.  This is necessary because JavaScript intercepts the normal form-submit action, and is responsible for sending all relevant data to to the `asset#create` action.  Also on the `asset#new` view, display a message to the user telling them which source the asset will be associated with upon creation.

* Update the JavaScript so that it sends the appropriate `source_id` data to the `asset#create` action.  The name of the function in which JavaScript posts to `asset#create` is `postAssetRecord`.

I have tested this locally in my browser, and it is working as expected.  I have also ensured that all Jasmine tests are passing.

### DIFF
--- a/app/controllers/audios_controller.rb
+++ b/app/controllers/audios_controller.rb
@@ -20,6 +20,7 @@ class AudiosController < ApplicationController
     @formdef = PSSBrowserUploads.av_form_definition('audio')
     @accepted_types = %w(.mp3 .mp2 .mp4a .wav .flac .aif .aiff .wma .mpga .oga
                          .ogg .au .adp .aac .weba).join(',')
+    @source = Source.find_by_id(params[:source_id])
   end
 
   ##
@@ -53,7 +54,11 @@ class AudiosController < ApplicationController
   private
 
   def audio_params
-    params.require(:audio).permit(:file_base,
-                                  source_ids: [])
+    ##
+    # Note that :source_ids is expressed as single hash, rather than
+    # "source_ids: []", as seen in other controllers.
+    # The single hash notation is required to work with the JavaScript code
+    # @see: app/javascripts/avupload.js
+    params.require(:audio).permit(:file_base, :source_ids)
   end
 end
--- a/app/assets/javascripts/avupload.js
+++ b/app/assets/javascripts/avupload.js
@@ -7,12 +7,16 @@ function AVUploadHandler() { UploadHandler.call(this); }
 AVUploadHandler.prototype = Object.create(UploadHandler.prototype);
 
 
-AVUploadHandler.prototype.createAssetRecord = function(key) {
+AVUploadHandler.prototype.createAssetRecord = function(postdata) {
     // For file basename, get rid of "video/" or "audio/" path part and
     // file extension.
-    file_base = key.replace(/^[a-z]+\/(.*)\.[a-z0-9]+$/i, "$1");
-    postdata = {};
-    postdata[pss_asset_type] = {file_base: file_base, key: key};
+    var key = postdata[pss_asset_type]['file_name'];
+    var file_base = key.replace(/^[a-z]+\/(.*)\.[a-z0-9]+$/i, "$1");
+
+    postdata[pss_asset_type]['file_base'] = file_base;
+    postdata[pss_asset_type]['key'] = key;
+    delete postdata[pss_asset_type]['file_name'];
+
     this.postAssetRecord(postdata);
 };
 
--- a/app/assets/javascripts/docupload.js
+++ b/app/assets/javascripts/docupload.js
@@ -7,9 +7,7 @@ function DocUploadHandler() { UploadHandler.call(this); }
 DocUploadHandler.prototype = Object.create(UploadHandler.prototype);
 
 
-DocUploadHandler.prototype.createAssetRecord = function(key) {
-    postdata = {};
-    postdata[pss_asset_type] = {file_name: key};
+DocUploadHandler.prototype.createAssetRecord = function(postdata) {
     this.postAssetRecord(postdata);
 };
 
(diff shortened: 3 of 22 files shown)


AudreyAltman:
[PR#81](https://github.com/dpla/primary-source-sets/pull/81) should be reviewed _before_ this.  PR#81 establishes some patterns for Jasmine logic that are used here, and if we need to re-think those patterns, it would be better to make those changes first to PR#81.

app/controllers/audios_controller.rb
```diff
@@ -20,6 +20,7 @@ def new
     @formdef = PSSBrowserUploads.av_form_definition('audio')
     @accepted_types = %w(.mp3 .mp2 .mp4a .wav .flac .aif .aiff .wma .mpga .oga
                          .ogg .au .adp .aac .weba).join(',')
+    @source = Source.find_by_id(params[:source_id])
```
markbreedlove: Could be that I'm missing something, but does `Source` need to be queried here, and in the other controllers? It seems that the goal is to pass the ID along to the Javascript that processes the form, so that it can put the source ID into the post data that it submits.

I could see querying Source here if the purpose is to verify that the ID exists; but then that check should be added.

markbreedlove:
There are some conditionals that are identical across the various `*UploadHandler` classes.  For example: https://github.com/dpla/primary-source-sets/blob/feature/8073-media-forms/app/assets/javascripts/avupload.js#L17-L25

Could these be factored out into a method of the `UploadHandler` base class, which is called by `UploadHandler.handleResponseIfDone()` to initialize a `postdata` object before passing it to `createAssetRecord`?  It would become necessary to pass `key` to this new function to initialize `postdata` instead of to `createAssetRecord`. That new function would have to set the `file_name` property of `postdata` with the value of `key`.

The call to a function that updates `postdata` could be inserted here: https://github.com/dpla/primary-source-sets/blob/feature/8073-media-forms/app/assets/javascripts/upload.js#L76

markbreedlove:
By the way, I don't know if this is the place to correct it, but there's another one of my lines where I have left out `var`:
https://github.com/dpla/primary-source-sets/blob/feature/8073-media-forms/app/assets/javascripts/upload.js#L75

markbreedlove:
Whatever the case, this works well, judging by what I see in my browser and in the database after adding an image.

app/controllers/audios_controller.rb
AudreyAltman: I instantiated the `Source` for two reasons.  First, as you noted, it works as check to verify that the ID exists.  The checks are in the views themselves.  See for instance: https://github.com/dpla/primary-source-sets/pull/82/files#diff-87342b07e87a296b812ab3e844a147efR6  But let me know if you think there's a better way to do the check.

Second, the views use other attributes of the `Source`, not just the ID.  See for instance: https://github.com/dpla/primary-source-sets/pull/82/files#diff-87342b07e87a296b812ab3e844a147efR16

AudreyAltman:
@markbreedlove this is ready for re-review.  See commit [87cbf99](https://github.com/dpla/primary-source-sets/commit/87cbf99d25d8646b7c8ad3441489d9fb1931e366) for the refactor of the javascript `*UploadHandler` classes.  I like your idea - let me know if I executed it how you imagined.  I also added in that `var` declaration that you suggested.

app/controllers/audios_controller.rb
markbreedlove: OK, that's cool.

It's out of scope here, but one thing that confuses me is:
```
[2] pry(main)> Source.find_by_id(99)
  Source Load (1.4ms)  SELECT  "sources".* FROM "sources"  WHERE "sources"."id" = 99 LIMIT 1
=> nil
[3] pry(main)> Source.find(99)
  Source Load (1.5ms)  SELECT  "sources".* FROM "sources"  WHERE "sources"."id" = ? LIMIT 1  [["id", 99]]
ActiveRecord::RecordNotFound: Couldn't find Source with 'id'=99
from /home/vagrant/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/activerecord-4.1.13/lib/active_record/relation/finder_methods.rb:320:in `raise_record_not_found_exception!'
[4] pry(main)> 
```
That inconsistency isn't what I'd expected, and I thought that `find_by_id` would also have thrown that exception.  It's another topic, though, not a problem here and now.

markbreedlove:
Sure, @AudreyAltman, looks good!  I'm ready to merge it, pending a squash.

AudreyAltman:
:squash:ed

PR #83, opened 2015-12-16
Refactor application controller

markbreedlove:
This refactors some methods out of `ApplicationController`, adding them to new `Encoder` and `NotificationEndpoint` concerns. The `ZencoderAuthentication` concern has been removed, and its authentication statement has been subsumed by `NotificationEndpoint`.

In the process of refactoring, `Encoder#create_media` has been given the ability to handle exceptions thrown by `#create_transcoding_job`, which had apparently been left out as an oversight.

Coverage should be slightly improved with the refactored tests.

I've removed an unnecessary test for `http_basic_authenticate_with` because that functionality should be verified by Rails's tests.

See https://issues.dp.la/issues/8072


app/controllers/concerns/encoder.rb
```diff
@@ -0,0 +1,136 @@
...
+    if model.save
+      render json: { id: model.id,
+                     resource: self.send("#{type}_path", model) },
+             status: :created
+    else
+      msg = "Could not save #{model.capitalize} record"
+      logger.error msg
+      render json: { message: msg }, status: :internal_server_error
+    end
+  end
```
AudreyAltman: :+1: on the improved error handling in this method.

app/controllers/concerns/encoder.rb
```diff
@@ -0,0 +1,136 @@
...
+  # @abstract Return the appropriate Settings array of hashes, i.e.
+  # video_outputs or audio_outputs.  These are the file extensions that will be
+  # generated.
+  #
+  def media_outputs
+    raise NotImplementedError
+  end
+
+
+  private
```
AudreyAltman: :+1: on making these methods private

app/controllers/concerns/notification_endpoint.rb
```diff
@@ -0,0 +1,33 @@
+
```
AudreyAltman: This module might benefit from top-level documentation explaining its main purpose.

AudreyAltman:
@markbreedlove This all looks good to me.  I made one suggestion of a place to add a comment, but otherwise, I think this is good to go, once its been rebased to develop, and squashed if you're so inclined.

markbreedlove:
OK, that's squashed into two commits and rebased.

PR #84, opened 2015-12-16
admin-only crud interfaces for tag, vocabulary

AudreyAltman:
This introduces admin-only CRUD interfaces for `tag` and `vocabulary`.  Using this data in the non-admin UI is out of scope for this PR.

According to the [relational database model](https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Relational+Database+Model), `tags` have many-to-many relationships with both `vocabularies` and `source_sets`.

This follows boilerplate CRUD patterns that have been established elsewhere in the application.  A couple of things to note:

* `Tag` and `Vocabulary` have "friendly" id slugs.

* `Tag` validates the uniqueness of the `label` attribute.  `Vocabulary` validates the uniqueness of the `name` attribute.  This is to avoid duplications, which would be confusing to an admin user.

* The `tag#new` controller accepts params for `source_set_id` and `vocabulary_id`.  This is to support admin workflows.  From `source_set#show`, an admin can click "Add new tag" and be taken to a `new` `tag` form that has pre-set the association between the `source_set` and the new `tag`.  Ditto if an admin starts from `vocabulary#show`.

This addresses [ticket #8103](https://issues.dp.la/issues/8103).

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1,0 +1,16 @@
+class Tag < ActiveRecord::Base
+  extend FriendlyId
+  has_and_belongs_to_many :source_sets
+  has_and_belongs_to_many :vocabularies
+  validates :label, presence: true, uniqueness: true
+
+  ##
+  # FriendlyId generates a human-readable slug to be used in the URL, in place
+  # of the ID.  The slug is automatically generated from the name field.
+  #
+  # Example:
+  #   Tag.create({ name: "Little My" }).slug = "little-my"
+  #   URL:  http://example.com/primary-source-sets/tags/little-my
+  #   To find this object: Tag.friendly.find("little-my")
+  friendly_id :label, use: :slugged
+end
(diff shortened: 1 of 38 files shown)


AudreyAltman:
I assigned this to @markbreedlove, but it could also be reviewed by @no-reply or @anarchivist.

markbreedlove:
This looks good to me. I have not run it, though. I'll assign it back to @AudreyAltman in case she wants to have someone else look at it, otherwise, I'm OK with it.

app/models/tag.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :label, presence: true, uniqueness: true
+
+  ##
+  # FriendlyId generates a human-readable slug to be used in the URL, in place
+  # of the ID.  The slug is automatically generated from the name field.
+  #
+  # Example:
+  #   Tag.create({ name: "Little My" }).slug = "little-my"
+  #   URL:  http://example.com/primary-source-sets/tags/little-my
+  #   To find this object: Tag.friendly.find("little-my")
```
anarchivist: little-my_slider-300x189

anarchivist:
:+1: :recycle: :baseball: :exclamation:

AudreyAltman:
Rebased to develop. Also, I have run this locally and can confirm that it operates as expected.

PR #85, opened 2015-12-21
persist year attribute for source_sets

AudreyAltman:
This persists a `year` attribute for `source_sets`.  It adds a column to the `source_sets` table.  The model validates that `year` is an integer.  It provides an interface for admins to add, view, and edit the `year`.  

This addresses [ticket #8204](https://issues.dp.la/issues/8204).

### DIFF
--- a/app/models/source_set.rb
+++ b/app/models/source_set.rb
@@ -7,6 +7,8 @@ class SourceSet < ActiveRecord::Base
   has_and_belongs_to_many :authors
   has_and_belongs_to_many :tags
   validates :name, presence: true
+  validates_numericality_of :year, only_integer: true, allow_nil: true,
+                                   less_than_or_equal_to: Date.today.year
 
   ##
   # FriendlyId generates a human-readable slug to be used in the URL, in place
--- a/app/controllers/source_sets_controller.rb
+++ b/app/controllers/source_sets_controller.rb
@@ -60,6 +60,7 @@ class SourceSetsController < ApplicationController
                                        :overview,
                                        :resources,
                                        :published,
+                                       :year,
                                        author_ids: [],
                                        tag_ids: [])
   end
--- a/app/views/source_sets/_form.html.erb
+++ b/app/views/source_sets/_form.html.erb
@@ -26,6 +26,13 @@
     <strong><%= f.label :published, style: 'display:inline' %></strong>
     <%= f.check_box :published %><br />
   </p>
+
+  <p>
+    <strong><%= f.label :year %></strong>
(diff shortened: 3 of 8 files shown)


app/models/source_set.rb
```diff
@@ -7,6 +7,7 @@ class SourceSet < ActiveRecord::Base
   has_and_belongs_to_many :authors
   has_and_belongs_to_many :tags
   validates :name, presence: true
+  validates_numericality_of :year, only_integer: true, allow_nil: true
```
markbreedlove: Is it worthwhile to check the range?

markbreedlove:
Is this year the date of creation of the source resource? Is it possible it would ever be BCE, for instance, for a picture of some sculpture from a museum?

Other than that, and my question about validating the range, I don't have any other observations, and this looks good to me.

AudreyAltman:
The year is a subject related to the items in the set.  It could be BCE, but its very unlikely.  If it were, I image that we could use a negative number to represent a BCE date.  This year is not going to be shown the end user, it's just going to be used to sort source_sets behind-the-scenes.  

It would probably be helpful to put at least a maximum cap on the date range, say 2100.  I'm not sure about a minimum cap because I can picture a world in which someone wants to make a set about fossils or something.

markbreedlove:
I guess that, in that case, it's a matter of documentation and the perceived needs of future admins. If the maintainers grows to a group larger than a few internal users, they might be prone to forget that the year is just a sort-order key; but I'm not sure how important this is to the project. Maybe a tooltip or label would help to clarify that it's just for sort order.

Another thing is that ranges like "2014-5" or phrases like "1920s" would be invalid, and I could see new  admins trying to enter those -- without a tooltip or label for guidance.

AudreyAltman:
@markbreedlove I added a [validation](https://github.com/dpla/primary-source-sets/pull/85/files#diff-8aea8a5bfdb4698272ac5b9ee26ba88fR11) and [test](https://github.com/dpla/primary-source-sets/pull/85/files#diff-33ec3c1eec2780823eb2e4cda3b73af1R47) to ensure that the year can't be greater in value than the current year.

AudreyAltman:
@markbreedlove I just saw your latest comment about adding instructions to the admin interface - doing that now.

AudreyAltman:
@markbreedlove instructions have been added to the view: https://github.com/dpla/primary-source-sets/pull/85/files#diff-8b5366e103d161b9136995d3c47ea222R32

markbreedlove:
Thanks! Good to go ...

PR #86, opened 2015-12-21
move source_name helper to source model

AudreyAltman:
This moves a helper method into a model.  The helper method was `source_name`.  It returned a `source`'s `name` if present; if not, it returned its `aggregation`.  This functionality makes more sense as an instance method of the `Source` model, so I moved it, renamed it `display_name`, and wrote a test.  I also changed the associated views and checked them all locally.

This addresses [ticket #8202](https://issues.dp.la/issues/8202).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,9 +1,5 @@
 module ApplicationHelper
 
-  def source_name(source)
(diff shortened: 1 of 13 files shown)


anarchivist:
:+1: Rebase?

AudreyAltman:
@anarchivist rebased to develop

PR #87, opened 2015-12-21
move author list from helper to model concern

AudreyAltman:
This refactors existing code by introducing a concern module for models with associated authors.  Functionality is factored out of the `ApplicationHelper` and the `Source` and `Guide` models.

This addresses [ticket #8202](https://issues.dp.la/issues/8202).

### DIFF
--- a/app/views/source_sets/show.html.erb
+++ b/app/views/source_sets/show.html.erb
@@ -24,7 +24,7 @@
         <h1><%= inline_markdown(@source_set.name) %></h1>
         <span class='subheading'>Primary Source Set</span>
         <% if @source_set.authors.present? %>
-          <span class='byline'>By <%= authors(@source_set).join(' and ') %></span>
+          <span class='byline'>By <%= @source_set.author_list.to_sentence %></span>
         <% end %>
       </div>
     </div>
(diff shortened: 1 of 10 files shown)


AudreyAltman:
This is assigned to @markbreedlove, but could be reviewed by @no-reply or @anarchivist instead.

app/views/source_sets/show.html.erb
```diff
@@ -24,7 +24,7 @@
         <h1><%= inline_markdown(@source_set.name) %></h1>
         <span class='subheading'>Primary Source Set</span>
         <% if @source_set.authors.present? %>
-          <span class='byline'>By <%= authors(@source_set).join(' and ') %></span>
+          <span class='byline'>By <%= @source_set.author_list.join(' and ') %></span>
```
anarchivist: Do we ever expect a set to have more than 2 authors? I'm just wondering how it'll look if we we have something like `Moomin Troll, Moominvalley Historical Society and Moominmama and Moominpapa` ...

AudreyAltman:
@anarchivist great point!  There may be a case with more than two authors.  I changed it from `join(' and ')` to `to_sentence`:

https://github.com/dpla/primary-source-sets/pull/87/files#diff-7068d5d33c70b367137656fdea6168a7R27

https://github.com/dpla/primary-source-sets/pull/87/files#diff-d18bc7e6b81d90f582b475cfa9d0c18aR26

anarchivist:
:+1:

PR #90, opened 2015-12-22
style form inputs and text areas

AudreyAltman:
This adds a little style to the admin input forms to make page elements more visible, thus making data entry easier.  I added a class name to the admin forms so that they can be distinguished from other forms on the pages, such as the DPLA search box.

This also corrects an oversight in the `tag` and `vocabulary` forms - to be consistent with the other admin forms in the application, they needed "Submit" buttons at the top of the form, as well as the bottom.

I have run the views locally and confirmed that they look as expected.

This addresses [ticket #8206](https://issues.dp.la/issues/8206).

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -205,7 +205,25 @@ input.form-submit {
   padding: 2px;
 }
 
-textarea {
-  height: 300px;
+.pss-admin-form input, #upload-form input.form-submit {
+  border: 1px solid grey;
+  padding: 0.5em;
+  margin-bottom: 0.5em;
+}
+
+.pss-admin-form input.form-submit, #upload-form input.form-submit {
+  background-color: #E6A425;
+}
+
+.pss-admin-form textarea {
+  height: 100px;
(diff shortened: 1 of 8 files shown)


markbreedlove:
Looks good.  There one form that I have noticed that lacks borders around `input`s is the "New Image" form -- screenshot attached.

In the "New set" view, the "Year" `input` is probably wider than it has to be -- screenshot also attached.

screen shot 2015-12-22 at 11 37 28 am
screen shot 2015-12-22 at 11 39 34 am

markbreedlove:
Oh, minor note on "New Image."  The height and width fields should be going away, pending work that's being done on the OpenSeadragon integration. I think the "alt text" field may need to stay.

AudreyAltman:
@markbreedlove good catches!  I've updated the CSS.  See screenshots below.

`image` form (the "upload" buttons look the same on `document`, `video`, and `audio`):

screen shot 2015-12-22 at 1 56 08 pm

`source_set` year box:

screen shot 2015-12-22 at 1 55 48 pm

markbreedlove:
OK, great!

PR #92, opened 2015-12-23
add specs that test if views render

AudreyAltman:
This adds rspec tests to all views confirming that they render without error.

Most views can be tested using the shared example `'renderable view'`.  However, the new media views (`audio#new`, `video#new`, `image#new`, and `document#new`) require a special shared example that stubs out the `s3form`.  This form is rendered through the `s3_browser_uploads` gem, and seems to require a fair bit of logic to test - so I am thinking that testing whether or not the `s3form` renders may be out of scope for this PR.

This also removes a statement from `spec_helper` that explicitly loads view specs.  I had added this back in [PR#21](https://github.com/dpla/primary-source-sets/pull/21) because view files hadn't been loading when we ran `rspec`.  But upon further local testing, it seems that this line is not needed, and is, in fact, causing the view specs to run twice.

This addresses [ticket #8208](https://issues.dp.la/issues/8208),

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -6,9 +6,6 @@ require File.expand_path('../../config/environment', __FILE__)
 # Load support files
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each { |f| require f }
 
-# Load view files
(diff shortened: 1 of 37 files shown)


anarchivist:
:+1: rebase?

AudreyAltman:
@anarchivist rebased.

PR #93, opened 2015-12-23
Replace image viewer with OpenSeadragon

markbreedlove:
Remove lightbox2 and add OpenSeadragon.

Remove the height and width fields from Image. OpenSeadragon loads images dynamically into a `<canvas>` element and the dimensions are not useful.

There are some minor changes to the stylesheets and the factory for Images, and the way that right-click downloads is prevented has been changed to use fewer lines of code.


app/views/images/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :head_script do %>
-<%= javascript_include_tag 'lightbox' %>
+<% javascript_include_tag 'openseadragon' %>
```
AudreyAltman: Could we put this in `:foot_script` instead?  This will keep it consistent with changes in PR#94.

app/views/sources/_image.html.erb
```diff
@@ -1,20 +1,15 @@
+<% content_for :head_script do %>
+<% javascript_include_tag 'openseadragon' %>
+<% end %>
```
AudreyAltman: Same as above:  Could we put this in :foot_script instead? This will keep it consistent with changes in PR#94.

app/views/images/show.html.erb
```diff
@@ -20,24 +20,14 @@
...
     <td><%= @image.alt_text %></td>
   </tr>
 </table>
 
-<a data-lightbox="image-1" data-title="<%= @image.alt_text %>"
-   href="<%= base_src + @image.file_name %>">
-   <img src="<%= base_src + @image.file_name %>" alt="<%= @image.alt_text %>"
-        style="max-width: 600px; max-height: 600px;" />
-</a>
+<div id="osd-viewer" style="width: 600px; height: 600px;"></div>
```
AudreyAltman: Could we put the style statements in a CSS file, like `primary_source_sets.css`?

app/views/sources/_image.html.erb
```diff
@@ -1,20 +1,15 @@
...
-         href="<%= base_src + @source.main_asset.file_name %>">
-
-        <div style="position: relative;">
-            <div style="position: absolute; top: 0; left: 0; width: 700px; height: 485px; z-index: 2;"></div>
-            <img src="<%= base_src + @source.main_asset.file_name %>"
-                 alt="<%= @source.main_asset.alt_text %>"
-                 style="position: absolute; top: 0; left: 0; z-index: 1; max-width: 700px; max-height: 485px;" />
-        </div>
+      <div id="osd-viewer" oncontextmenu="return false;"></div>
+      <p id="image-caption"><%= html_escape_once(@source.main_asset.alt_text) %></p>
```
AudreyAltman: I'm not familiar with `html_escape_once`.  What's the benefit of using it instead of `html_escape`?

app/assets/stylesheets/primary_source_sets.css
```diff
@@ -18,6 +18,12 @@
   margin-top: 2em;
 }
 
+#osd-viewer {
+  width: 700px;
+  height: 485px;
+  background-color: black;
+}
```
AudreyAltman: I noticed that on the `source#show` page, the viewer isn't taking up the whole space to the left of the sidebar.  I tried changing the style (above) to `width: 100%; height: 100%` and that worked for me.

AudreyAltman:
I've made a few comments, but all in all this is looking really good.

app/views/sources/_image.html.erb
markbreedlove: From what I've quickly read, it's supposed to HTML-escape a string, but should skip character sequences that are already encoded.  For example, if someone is being diligent and entering `this &amp; that`, it will not further encode the ampersand in that string.

app/views/images/show.html.erb
markbreedlove: Sure, thanks.  This was an oversight.

It reminds me, though, that the dimensions for the viewer in images/show were different than those in sources/show, so I may need to make some adjustments -- or maybe not; I'll see.

It's good to take these styles out of the HTML so that we can eventually get the app to display properly on a mobile device with alternate media stylesheets.

app/views/images/show.html.erb
markbreedlove: I've just learned something new.  Most browsers now support `async` and `defer` attributes for the `script` element, per http://stackoverflow.com/questions/436411/where-is-the-best-place-to-put-script-tags-in-html-markup, so I'm looking into whether I can do that with `javascript_include_tag`.

AudreyAltman: Looks promising!

app/views/sources/_image.html.erb
AudreyAltman: :+1:

app/assets/stylesheets/primary_source_sets.css
markbreedlove: Sure, sounds good.

markbreedlove:
OK, @AudreyAltman, I think that's probably it. Do you think I should squash this?

PR #94, opened 2015-12-23
Fix/move javascript

AudreyAltman:
This moves javascripts to the bottom of rendered HTML pages in order to ensure proper loading.  It also moves the `foot_script` (the section of the HTML where the javascripts go) _after_ the `</body>` tag in the HTML, rather than before.

I've tested this locally and all javascripts seem to still be working as expected.

This addresses [ticket #8067](https://issues.dp.la/issues/8067).

### DIFF
--- a/app/views/audios/new.html.erb
+++ b/app/views/audios/new.html.erb
@@ -1,10 +1,12 @@
-<% content_for :head_script do %>
+<% content_for :foot_script do %>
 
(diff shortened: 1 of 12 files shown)


anarchivist:
:+1: rebase?

AudreyAltman:
@anarchivist rebased

PR #96, opened 2016-01-04
Fix double-encoding of characters in titles

markbreedlove:
Remove the double-HTML-encoding of characters into HTML entities.

See https://issues.dp.la/issues/8199

The `#inline_markdown` method call is not necessary for `<title>` elements, because it's not desirable to have markup in titles -- all we need is straight text.  The `#inline_markdown` method encodes HTML entities, and the resulting string was being encoded one more time by Rails automatically when the view was rendered.

### DIFF
--- a/app/helpers/markdown_helper.rb
+++ b/app/helpers/markdown_helper.rb
@@ -1,3 +1,5 @@
+require 'redcarpet/render_strip'
+
 module MarkdownHelper
   ##
   # @param String text A markdown formatted String
@@ -16,6 +18,10 @@ module MarkdownHelper
   # @raise TypeError if @param is not a String
   def inline_markdown(text)
     strip_p_tags(markdown(text))
+  end
+
+  def plaintext_from_md(text)
+    Redcarpet::Markdown.new(Redcarpet::Render::StripDown).render(text)
   end
 
   private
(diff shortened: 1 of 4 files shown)


anarchivist:
> The `#inline_markdown` method call is not necessary for `<title>` elements, because it's not desirable to have markup in titles -- all we need is straight text.  

I think the issue is that set titles themselves will occasionally contain Markdown (see, for example, the set about [*A Raisin in the Sun*](http://dp.la/primary-source-sets/sets/a-raisin-in-the-sun-by-lorraine-hansberry) - the title is set to `*A Raisin in the Sun* by Lorraine Hansberry`). My guess is we also need something that will strip Markdown from the title data when it gets passed into the `<title/>` element. 

Maybe we could use a different `redcarpet` renderer class, like [`Redcarpet::Render::StripDown`](http://www.rubydoc.info/gems/redcarpet/3.2.2/Redcarpet/Render/StripDown)?

AudreyAltman:
Yes, @anarchivist has it right.  The idea was to strip the markdown characters (such as `*`) out of the titles in a two-step process:  First, render the HTML with `inline_markdown`.  Second, strip out the HTML elements with `strip_tags`.  This would, in theory, leave us with plain text.  But it looks like `Redcarpet::Render::StripDown` might be a more elegant (and less buggy) solution.

markbreedlove:
OK, @AudreyAltman and @anarchivist, that last change brings in `Redcarpet::Render::StripDown`.

No HTML tags are stripped, but they'll get turned into HTML entities. The assumption is that the maintainers are supposed to know to use Markup, not HTML, and things will look funny anyway if they use HTML markup in their text.

AudreyAltman:
:+1: :squash:?

anarchivist:
:ok_hand:

markbreedlove:
OK, that should be ready to merge.

PR #97, opened 2016-01-04
Fix "404 Not Found" errors with assets

markbreedlove:
Though the `frontend-assets` gem is retained, get the Source Sans Pro font from Google Fonts. (e.g. https://www.google.com/fonts#QuickUsePlace:quickUse/Family:Source+Sans+Pro:400,600,400italic,600italic).  This fixes the loading of the fonts in development, avoids unnecessary complexity, and probably makes the pages load faster by parallelizing browser requests across domains.

Fix the broken links to the magnifying glass and background images that are included in the CSS by using the Rails asset pipeline functions to include the images regardless of whether they're in their development or production locations.

References https://issues.dp.la/issues/8090

Note that I added a couple of "FIXME" comments regarding missing images that are included in the CSS. I think that there are more that I didn't flag.


AudreyAltman:
This looks good.  I like the use of the Google fonts.  

Its worth noting that by changing `app/assets/stylesheets/vendor/main.css.erb`, the file is no longer identical its counterpart in the `frontend` app.  I think that's okay - hopefully we will address this and other redundancies (or near-redundancies) when we refactor code that is shared amongst the various frontend applications.

:squash: and :merge:?

markbreedlove:
OK, squashed.

AudreyAltman:
rebase?

markbreedlove:
OK, rebased and CI passes.

PR #99, opened 2016-01-11
show and filter by tags on user interface

AudreyAltman:
This adds functionality to show `tags` on the user interface, and allows users to filter by `tags`.  `tags` only appear on the the `source_sets#index` view.

Basic CRUD interfaces for tags were added in [PR#84](https://github.com/dpla/primary-source-sets/pull/84); however, tags were _only_ visible to logged-in admins.  This PR introduces functionality available to non-logged-in users.

Here is an example of the how the functionality should work:

A user goes to `/primary-source-sets/sets?tags[]=dogs` and sees only those sets with a tag whose label is `dogs`.  All of the tags associated with a set (including `dogs`) are listed beneath each set.

One set also has a tag `cats`.  The user clicks on `cats`, and is taken to `/primary-source-sets/sets?tags[]=dogs&tags[]=cats`.  The users is shown all sets having _both_ `cats` and `dogs` tags.

Changes include:

* Updating the `source_sets#index` controller to parse a `tags` param.  

* Adding methods the `SourceSets` model to request sets by tag.

* Updating the views to show the tags and allow filtering.

* Styling, including changes to the HTML and CSS.  This includes a JavaScript function to ensure that that the dark blue boxes containing the name of each set are the same height in each row.

* A system config that allows us specify whether or not to display tags to the end user.  This will help us to manage deployment - admins need time to add all of the tags before we start showing them to the end user.

### DIFF
--- a/app/models/source_set.rb
+++ b/app/models/source_set.rb
@@ -9,6 +9,7 @@ class SourceSet < ActiveRecord::Base
   validates :name, presence: true
   validates_numericality_of :year, only_integer: true, allow_nil: true,
                                    less_than_or_equal_to: Date.today.year
+  default_scope { order(created_at: :asc) }
 
   ##
   # FriendlyId generates a human-readable slug to be used in the URL, in place
@@ -31,4 +32,24 @@ class SourceSet < ActiveRecord::Base
   def self.unpublished_sets
     self.where(published: false)
   end
+
+  ##
+  # Get SourceSets associated with all of the specified tags.
+  # EACH returned SourceSet will have ALL of the tags.
+  # If no tags are specified, return all SourceSets.
+  # @param tags [Array<String>]
+  # @return [Array<SourceSet>]
+  def self.with_tags(tags)
+    return all unless tags.present?
+    tags.map { |tag| with_tag(tag) }.inject(:&)
+  end
+
+  ##
+  # Get SourceSets associated with the specified tag.
+  # @param tag [String]
+  # @return [SourceSet]
+  def self.with_tag(tag)
+    joins(:tags).where('tags.label = ?', tag)
+  end
+  private_class_method :with_tag
 end
--- a/app/assets/javascripts/style.js
+++ b/app/assets/javascripts/style.js
@@ -5,10 +5,12 @@
 
   $(document).ready(function() {
     setAsideHeight();
+    setNameContainerHeight();
   });
 
   $(window).resize(function() {
     setAsideHeight();
+    setNameContainerHeight();
   })
 })();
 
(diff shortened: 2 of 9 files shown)


app/models/source_set.rb
```diff
@@ -9,6 +9,7 @@ class SourceSet < ActiveRecord::Base
   validates :name, presence: true
   validates_numericality_of :year, only_integer: true, allow_nil: true,
                                    less_than_or_equal_to: Date.today.year
+  default_scope { order(created_at: :asc) }
```
AudreyAltman: I had to move this out of the [view](https://github.com/dpla/primary-source-sets/pull/99/files#diff-a1380fab38a891ab0e8d364162e964caL2).

markbreedlove:
Thanks, @AudreyAltman! It works nicely.

General questions:
* Could the `enable_tags` setting be renamed to clarify that it's for enabling the display of tags on the front-end?
* Can you explain what vocabularies are? (Is this something that needs a tooltip or label within the application?)
* I have a hunch that there's some pure CSS way to do some of the dimensional manipulations, or to design around having to do them, but it's probably not worth figuring out at the moment.

Re. `/tags/new`:
* What is the tag's URI for? (It lets me enter an invalid, badly-formed URI, e.g. "not a URI")
* The "Vocabularies" label is apparently for a `#tag_vocabularies` input, which does not exist. There's `<input name="tag[vocabulary_ids][]">` but it has no id. This is with or without my having any Vocabularies entered.

Re `style.js`:
* If `$(this)` can be assigned to a variable in the various scopes in which it's used, it might help to document what it represents each time. Maybe not each time; maybe just in the outermost scope.

AudreyAltman:
Thanks for the feedback, @markbreedlove!  

I've changed the name of the `enable_tags` setting to `display_tags_on_public_ui`.  After Franky has had a chance to add all the tags, we may consider removing this setting - I don't really see an ongoing need for it, and it might be needless clutter.

I've also assigned a variable name for the outer `($this)` in `style.js`, along with a top-level comment for the function. Let me know if it need further clarification.

I made a ticket to capture your questions about the tags/vocabs interfaces: https://issues.dp.la/issues/8229  They seemed substantive enough for their own PR.

AudreyAltman:
On last thing - I would love to find a pure CSS way to do dimensional manipulations, but far I haven't had any luck.  I agree that it may be wise to put off that work until after our looming deadline.  I'll make a ticket about that, too.

AudreyAltman:
I've just thought of one last thing I need to do on the public user view, so please don't merge yet.

AudreyAltman:
Ok, I made the changes to the public user view.  They apply styles to unpublished sets.  You can see the diff here: https://github.com/dpla/primary-source-sets/commit/fa6f2667897d45bd205c7513aea7cea43919490d

markbreedlove:
OK, that all looks good! Should I merge it now?

AudreyAltman:
I will :squash: first.

AudreyAltman:
:squash:ed

PR #100, opened 2016-01-12
Fix mobile layout and menus

markbreedlove:
Fixes https://issues.dp.la/issues/8093

* Add "viewport" `<meta>` element to the layout that makes mobile browsers show the page at the correct scale.
* Add a "charset" `<meta>` element while we're at it, as a best practice.
* Enable the menu buttons that show the navigation menu when the screen is small.

It would be nice not to need code to pull off the menus. I've implemented an adaptive menu like this (with the menu that gets replaced by a button) in pure CSS, not that I understand it that well. We can probably get rid of some of the code in a future iteration. It's harder to do it differently now without rewriting a lot of stylesheets and templates.

### DIFF
--- a/app/assets/stylesheets/vendor/main.css.erb
+++ b/app/assets/stylesheets/vendor/main.css.erb
@@ -711,7 +711,7 @@ aside .module span { color: #000; }
     transition: 0.3s ease;
 }
 
-.btn:hover, .btn a:hover, .shareSave .btn a:hover, .menu-btn:hover, .toggle:hover {
+.btn:hover, .btn a:hover, .shareSave .btn a:hover, .toggle:hover {
     background: black;
     -webkit-transition: 0.3s ease;
     -moz-transition: 0.3s ease;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/vendor/main.css.erb
```diff
@@ -711,7 +711,7 @@ aside .module span { color: #000; }
     transition: 0.3s ease;
 }
 
-.btn:hover, .btn a:hover, .shareSave .btn a:hover, .menu-btn:hover, .toggle:hover {
+.btn:hover, .btn a:hover, .shareSave .btn a:hover, .toggle:hover {
```
AudreyAltman: It might be good to keep the `.menu-btn:hover` to maintain consistency with the rest of the website.  If you're in a desktop browser and resize your window so its really skinny, you would have use of the `:hover` state on this button.  Unless there's another reason for getting rid of it?

markbreedlove: I got rid of it because, on a mobile device, the "hover" state stays activated after you click on the link. It turns black and doesn't turn orange again. There may be some better way of handling it, but I was trying not to spend too much time on it.

AudreyAltman: That makes sense.  In that case, I agree that this is a good fix.

AudreyAltman:
I think this looks great.  Do you want to :squash: before I :merge:?

markbreedlove:
That should be ready to go now.

PR #101, opened 2016-01-12
order sets by date created or year

AudreyAltman:
This introduces a feature that allows users to sort sets by `created_date` and `year` on the `sets#index` view.  

The sort option is included in the URI as an `order` param.  For example: `/sets?order=recently_added'`  The application recognizes three legitimate `order` params, `recently_added`, `chronology_asc`, and `chronology_desc`.  Any other `order` param is ignored.  If there is no legitimate `order` param, it defaults to `recently_added`.

Changes include:

* A method in the `SourceSet` model that parses the `order` param and adds sort options to a database query.

* Updating the `source_sets#index` controller method.

* Adding a dropdown menu to the view so users can select a sort option.

* Unobtrusive JavaScript.  If JavaScript is disabled, a form submit button will appear.  If JavaScript is enabled, the form submit button will be hidden and the dropdown menu will auto-submit upon change.

This addresses [ticket #8205](https://issues.dp.la/issues/8205).

### DIFF
--- a/app/models/source_set.rb
+++ b/app/models/source_set.rb
@@ -9,7 +9,6 @@ class SourceSet < ActiveRecord::Base
   validates :name, presence: true
   validates_numericality_of :year, only_integer: true, allow_nil: true,
                                    less_than_or_equal_to: Date.today.year
-  default_scope { order(created_at: :asc) }
 
   ##
   # FriendlyId generates a human-readable slug to be used in the URL, in place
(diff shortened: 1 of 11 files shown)


app/models/source_set.rb
```diff
@@ -34,14 +33,27 @@ def self.unpublished_sets
...
   # Get SourceSets associated with all of the specified tags.
   # EACH returned SourceSet will have ALL of the tags.
   # If no tags are specified, return all SourceSets.
-  # @param tags [Array<String>]
+  # @param tags [Array<String>] or nil
   # @return [Array<SourceSet>]
   def self.with_tags(tags)
     return all unless tags.present?
-    tags.map { |tag| with_tag(tag) }.inject(:&)
+    tags.map { |tag| with_tag(tag) }.inject(:&).to_a
```
AudreyAltman: This is just correcting a bug I found.  Without `to_a` the method may return either an `Array` or an `ActiveRecord::Relation`.  This way, it always returns an `Array`.

AudreyAltman:
I've noticed in my local browser that using JavaScript to hide page elements after they've loaded is causing them to blink on the page.  I'm trying to troubleshoot this.

AudreyAltman:
Solved the blinking problem!  I enclosed the submit button in a `<noscript>` tag, so that it will not render if the browser does not have JavaScript enabled.

markbreedlove:
Looks great.

The only thing I wonder about is whether `source_sets#index` should ignore a bad `order` value. I'm usually nervous about silent failures. If this were an API endpoint I'd probably be more concerned.

In the rare event that someone fiddles with the querystring on purpose, it would be valuable to have a message printed to the page that says the value "foo" for "order" is not valid. On the other hand, if we change the valid values of the "order" parameter, and there are links pointing to the site, we don't want everyone who lands on the page to get an error message. I guess it's best to ignore invalid values for now, with reservations about doing so generally.

markbreedlove:
Also forgot to mention, it looks good and works well, both with and without Javascript enabled.

PR #102, opened 2016-01-13
fix settings file and muliple tags/orders params

AudreyAltman:
This solves two bugs that that were introduced in release v1.2.2.  

1. In v1.2.2, user-selected tags and orders are over-riding one another. For example, from the browser window, if I select a tag and then select and order, I will lose the tag definition in the URI, and visa versa. This ensures that when tags are selected/deselected, and when orders are changed, all relevant params are passed along.

2. It changes default value of `display_tags_on_public_ui` to `false`.


markbreedlove:
OK, that's clear ...

PR #104, opened 2016-01-15
Add account management

markbreedlove:
Add account management, mostly using Devise, and overriding some of its default behavior.

Accounts are only created by admins. When the account is created, an email is sent to the new account's address for confirmation. During confirmation, the new user must set a password.

README.md contains some notes about setting up `mailcatcher` for development with confirmation emails.

References https://issues.dp.la/issues/8162

Possible TODO:  Add a full-name field to Admin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,25 @@ it to "use local source," in other words, use this working copy as the source of
 deployments, please be aware that the working copy's configuration files will be
 copied into the `automation` VM, by design.
 
+### ActionMailer setup
+
+This project uses ActionMailer to send account-creation confirmation emails. We
+use the [mailcatcher](http://mailcatcher.me/) gem in development to simulate a
+mail server, rather than running a real one. Per [its
+documentation](http://mailcatcher.me/), we do not include it in our Gemfile, and
+you need to install it with `gem install mailcatcher`.
+
+The mail settings for production are ActionMailer `smtp_settings` key/value
+pairs.  There are defaults in `settings.yml` that can be overridden in
+`settings.local.yml`.
+
+If you're running in the development VM (with our Vagrantfile), you'll want to
+invoke `mailcatcher` this way:
+```
+mailcatcher --http-ip 0.0.0.0
+```
+This ensures that the HTTP interface will be listening on the network interface
+that is being forwarded to `localhost` on your host machine.
 
 Testing
 -------
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -1,0 +1,7 @@
+<p>Welcome, <%= @email %>!</p>
+
+<p>You have a new account for DPLA Primary Source Sets.</p>
+
+<p>You can confirm your account email through the link below:</p>
+
+<p><%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>
--- a/app/views/devise/passwords/new.html.erb
+++ b/app/views/devise/passwords/new.html.erb
@@ -1,0 +1,19 @@
+<h2>Forgot your password?</h2>
+
+<%= form_for(resource,
+             as: resource_name,
+             url: password_path(resource_name),
+             html: { method: :post, class: 'pss-admin-form' }) do |f| %>
+  <%= devise_error_messages! %>
+
+  <div class="field">
+    <%= f.label :email %><br />
+    <%= f.email_field :email, autofocus: true %>
+  </div>
+
+  <div class="actions">
+    <%= f.submit "Send me reset password instructions", class: 'form-submit' %>
+  </div>
+<% end %>
+
+<%= render "devise/shared/links" %>
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -1,0 +1,15 @@
+<%- if controller_name != 'sessions' %>
+  <%= link_to "Log in", new_session_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.recoverable? && controller_name != 'passwords' && controller_name != 'registrations' %>
+  <%= link_to "Forgot your password?", new_password_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
+  <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.lockable? && resource_class.unlock_strategy_enabled?(:email) && controller_name != 'unlocks' %>
+  <%= link_to "Didn't receive unlock instructions?", new_unlock_path(resource_name) %><br />
+<% end -%>
--- a/app/views/registrations/edit.html.erb
+++ b/app/views/registrations/edit.html.erb
@@ -1,0 +1,34 @@
+<h2>Edit Account</h2>
+
+<% if devise_mapping.confirmable? && resource.pending_reconfirmation? %>
+  <div>Currently awaiting confirmation for: <%= resource.unconfirmed_email %>
+  </div>
+<% end %>
+
+<%= devise_error_messages! %>
+
+<%# View for Devise method inherited by Registrations controller. %>
+<%# Uses `resource` for consistency with Devise::Registrations#new %>
+<%= form_for(resource,
+             as: resource_name,
+             url: registration_path(resource),
+             html: { class: 'pss-admin-form' }) do |f| %>
+
+  <%= f.label :email %><br />
+  <%= f.email_field :email, autofocus: true %>
+
+  <%= f.label :password %>
+  <%= f.password_field :password, autocomplete: "off" %>
+  <div><i>(leave blank if you don't want to change it)</i></div>
+
+  <%= f.label :password_confirmation %><br />
+  <%= f.password_field :password_confirmation, autocomplete: "off" %>
+
+  <div>
+    <%= f.submit "Update", class: 'form-submit' %>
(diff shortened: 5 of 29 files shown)


markbreedlove:
Uh-oh ... just noticed tests have been failing.  I'll address that, but review could probably start now.

markbreedlove:
Those tests have been fixed, and I've added some for the `Admin` model.

markbreedlove:
I have some changes I want to make to the registrations edit form (where you edit the account email and password) but I'll present those in an upcoming PR.

markbreedlove:
Actually, there's one change that has to happen here. There's a problem with `registrations#update` (i.e. Account Edit).  @AudreyAltman you might want to hold off looking at that.

markbreedlove:
@AudreyAltman should be OK now, though I acknowledge that tests are lacking for the overridden RegistrationsController methods.

db/migrate/20160113221031_add_confirmable_to_admin.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    Admin.update_all({
+      :confirmed_at => DateTime.now,
+      :confirmation_sent_at => DateTime.now
+    })
+  end
+
+  def self.down
+    remove_column :admins,
+                  [:confirmed_at, :confirmation_token, :confirmation_sent_at]
```
AudreyAltman: Why don't you want to remove `:unconfirmed_email` or `confirmation_token` index in the `down` function?

app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -0,0 +1,5 @@
+<p>Welcome <%= @email %>!</p>
+
```
AudreyAltman: Maybe add sentence here that says something like "You have a new account for DPLA Primary Source Sets" so that people know what the account is for.

db/migrate/20160113221031_add_confirmable_to_admin.rb
markbreedlove: Just an oversight -- will fix it.  Thanks!

app/views/registrations/new.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <%= f.email_field :email, autofocus: true %>
+
+  <%= f.label :password %>
+  <%= f.password_field :password, autocomplete: "off" %>
+  <% if @validatable %>
+  <div><em>(<%= @minimum_password_length %> characters minimum)</em></div>
+  <% end %>
+
+  <%= f.label :password_confirmation %>
+  <%= f.password_field :password_confirmation, autocomplete: "off" %>
```
AudreyAltman: Can we remove the ability to add a password to this form?  

I found that if I create a new account with a password, it gets really confusing for the new user:  The confirmation email is sent to the new email address.  Then, from the email, when I click "Confirm my account", I am taken, already signed in, to `primary-source-sets/sets`.  I never learn my password, so if I get logged out before I change it, I can't get back in again, and my confirmation token expires.

Plus, I don't think people will want to admins create new accounts and set passwords for those accounts, correct?

markbreedlove: I thought I did remove that already, but I could be confused.  Let me check; may have committed it elsewhere.

I intended for there to be no password form on the create page.

AudreyAltman:
I suppose we could delete the views that have to do with locking and unlocking b/c we haven't activated locking - but maybe good to keep them in case we want to activate locking at some point.

app/views/registrations/index.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1>Admins</h1>
+
+<p><%= link_to "Create new admin", new_registration_path %></p>
+
+<table>
+<% @admins.each do |admin| %>
+  <tr>
+    <td><%= admin.email %></td>
+    <td><%= link_to "Edit", edit_registration_path(admin) %></td>
```
AudreyAltman: Can we only provide an edit link to the logged-in admin's account?  I don't think we want admins editing one another's accounts.

app/views/registrations/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+  <%= f.label :password_confirmation %><br />
+  <%= f.password_field :password_confirmation, autocomplete: "off" %>
+
+  <div>
+    <%= f.submit "Update", class: 'form-submit' %>
+  </div>
+
+<% end %>
+
```
AudreyAltman: Should this form render if the user is not logged into the specific account that this form will edit?  Maybe render an error message or do a redirect if a logged in admin tries to access the edit form for someone else's account?

app/views/devise/passwords/new.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <%= devise_error_messages! %>
+
+  <div class="field">
+    <%= f.label :email %><br />
+    <%= f.email_field :email, autofocus: true %>
+  </div>
+
+  <div class="actions">
+    <%= f.submit "Send me reset password instructions" %>
+  </div>
```
AudreyAltman: CSS is hijacking your submit button and causing to render as a second magnifying glass by the search box.

screen shot 2016-01-22 at 11 16 34 am

AudreyAltman: Actually, I think the submit buttons on _all_ of the forms are rendering as magnifying glasses.

README.md
```diff
@@ -187,6 +187,25 @@ it to "use local source," in other words, use this working copy as the source of
...
+pairs.  There are defaults in `settings.yml` that can be overridden in
+`settings.local.yml`.
+
+If you're running in the development VM (with our Vagrantfile), you'll want to
+invoke `mailcatcher` this way:
+```
+mailcatcher --http-ip 0.0.0.0
+```
+This ensures that the HTTP interface will be listening on the network interface
+that is being forwarded to `localhost` on your host machine.
```
AudreyAltman: Mailcatcher is a great tool!  I've never used it before, but it's certainly great for testing.

app/views/devise/shared/_links.html.erb
```diff
@@ -0,0 +1,25 @@
+<%- if controller_name != 'sessions' %>
+  <%= link_to "Log in", new_session_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.registerable? && controller_name != 'registrations' %>
+  <%= link_to "Sign up", new_registration_path(resource_name) %><br />
+<% end -%>
```
AudreyAltman: Can we delete the "Sign up" link since we're not allowing people to sign up?

AudreyAltman:
Overall, this is great!  Despite the difficulties with Devise, what you've written fits our needs perfectly.  I appreciate the detailed comments.
  
In my browser, I’ve tested creating, editing, and deleting accounts, deleting an account I’m currently logged in as, changing password, resetting password if forgotten, and claiming I didn’t receive confirmation instructions.  Everything is working.

I've made a few comments, mostly things we could change in the views to help workflow.

db/migrate/20160113221031_add_confirmable_to_admin.rb
markbreedlove: OK, should be fixed.  `confirmation_token` was already being removed, and its index (which you may have been referring to in your comment) should be removed when the column is dropped.

app/views/devise/mailer/confirmation_instructions.html.erb
markbreedlove: OK.  We might add a ticket at some point to parameterize the name of the site.  It's hardcoded as *DPLA* Primary Source Sets

app/views/registrations/new.html.erb
markbreedlove: I must have been thinking of the edit page. Thanks for catching this.  I've removed the password fields. I'm having trouble getting the controller to reject password fields in `#create` because it's hard to override Devise's functionality (you can add allowed fields but it's not clear how to subtract them).  I don't think it's too much of a problem because nobody's going to try to fill in that field outside of a form submission, but I could look into it later.

markbreedlove:
Re https://github.com/dpla/primary-source-sets/pull/104#issuecomment-173961463, yes, I left those generated views in there, though I haven't thought a lot about the locking feature.  I think Devise locks the account after a certain number of failed login attempts.

app/views/registrations/index.html.erb
markbreedlove: Coming soon! That's a feature in the upcoming PR that I'm going to issue after this one is merged, when permissions and roles are introduced.

app/views/registrations/edit.html.erb
markbreedlove: That's another feature of the upcoming PR.

app/views/devise/passwords/new.html.erb
markbreedlove: Interesting. This is only happening for me on the login-related pages -- sign_in, password/new, confirmation/new ...

app/views/devise/shared/_links.html.erb
markbreedlove: OK, thanks! It's gone now.

app/views/devise/passwords/new.html.erb
markbreedlove: OK, I think I fixed the buttons.

The forms should have the CSS class 'pss-admin-form' and the submit buttons should have the class 'form-submit', and they didn't.

I've also removed one view that's unused and the existence of which was confusing.

markbreedlove:
Re account locking, @AudreyAltman, you are correct to note that we don't have the `:lockable` behavior enabled on the Admin model.  Not sure if it's an issue for this PR or something to follow up with next.

AudreyAltman:
Looks great.  :squash:?

PR #105, opened 2016-01-19
vocab-sensitive tag filtering

AudreyAltman:
This introduces a feature that allows public users to filter source sets by tags on the `source_sets#index` view.

There are two ways to filter - dropdown menus, and clicking on a linked tag listed under a set.

Changes to the code include:

* In the `source_sets#index` controller, the instance variable `@tags` used to be an Array of tag slugs, taken from `params[:tags]`.  However, the view need access to the entire Tag model, not just the slugs.  So, the value of `@tags` is now an Array of `Tag`s.

* There is a new boolean data field for `vocabularies` called `filter`.  This field allows the admin to determine whether or not a vocabulary will be filterable on public UI. It will help admin workflows, b/c they can ensure that all necessary data is in the database before “turning on” the filter.  Note that views only display data from “filterterable” `vocabularies` (i.e. `vocabularies` where `filter-true`).

* There is a helper method, `filters_for_sets`, which returns filterable `vocabularies` and their associated `tags` for a given group of sets (for public users, the group of sets will be published sets; for admins, it will be all sets).  By collecting all of this data into a Hash that is available to the view, I was able to significantly reduce calls to the database.

* There is an edge case in which a user keys in a URI that contains multiple tags from the same vocabulary.  This doesn’t happen naturally on the site, but could happen.  In that case, the `select` element appears as an HTML multi-select, ie. an expanded select menu with multiple terms hilighted.  Because the multi-select menu takes up so much space on the screen, it is not the default.

This PR may benefit from some refactoring, and there is one `rspec` test that is not yet working; however, in the interest of getting it out the door on time, I’m submitting it as in with the intention to iterate later.

### DIFF
--- a/app/controllers/source_sets_controller.rb
+++ b/app/controllers/source_sets_controller.rb
@@ -8,7 +8,7 @@ class SourceSetsController < ApplicationController
   before_action :authenticate_admin!, only: [:new, :edit]
 
   def index
-    @tags = params[:tags]
+    @tags = get_tags_from_params
     @order = params[:order]
     @published_sets = SourceSet.published_sets.order_by(@order).with_tags(@tags)
     @unpublished_sets = SourceSet.unpublished_sets.order_by(@order)
@@ -66,5 +66,24 @@ class SourceSetsController < ApplicationController
                                        :year,
                                        author_ids: [],
                                        tag_ids: [])
+  end
+
+  ##
+  # @return [Array<Tag>]
+  def get_tags_from_params
+    return nil unless params[:tags].present?
+    Tag.where("slug IN (?)", valid_tags_params)
+  end
+
+  ##
+  # Get only those :tags params whose characters are solely comprised of
+  # letters, numbers or '-'.
+  #
+  # @return [Array<String>]
+  def valid_tags_params
+    return [] unless params[:tags].present?
+    params[:tags].select do |slug|
+      (slug =~ /[^a-zA-Z0-9-]/).nil?
+    end
   end
 end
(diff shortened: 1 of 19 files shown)


app/controllers/source_sets_controller.rb
```diff
@@ -67,4 +67,11 @@ def source_set_params
                                        author_ids: [],
                                        tag_ids: [])
   end
+
+  ##
+  # @return [Array<Tag>]
+  def get_tags_from_params
+    return nil unless params[:tags].present?
+    params[:tags].map { |slug| Tag.find_by_slug(slug) }.compact
```
markbreedlove: Is there a way to have the query work on a list of values, instead of running individually for each value?

For example, is this sane?: `Tag.where("slug IN (?)", params[:tags])`

It's probably not a big deal with a small list of tags, but I'm trying to think of ways to reduce the number of database queries.

markbreedlove: One issue with my suggestion is it might require some input sanitization if that's not done automatically.

AudreyAltman:
@markbreedlove I streamlined and sanitized the query: https://github.com/dpla/primary-source-sets/commit/e317be1f35e293b208e093f1e16a1d00b2a2b4ec

markbreedlove:
OK, looks good!

AudreyAltman:
rebased - should be ready to :merge: if @markbreedlove is ready

AudreyAltman:
On second thought @markbreedlove I have some style changes to the UI I can merge in before we :merge:.  I'll do that now...

AudreyAltman:
Okay @markbreedlove I changed the UI display based on feedback from Franky.  It required adding a new helper method, for which there is also a test.  I've tested my local instance in browser, and everything is working as expected.

PR #107, opened 2016-01-20
link Primary Source Sets header

AudreyAltman:
On the `source_sets#index view`, link the "Primary Source Sets" header to `source_sets#index` without any order/filter parameters. This will serve as an "out" for users who become confused by the order/filter options. 

This feature was requested by Franky.  It addresses [ticket #8242](https://issues.dp.la/issues/8242).

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -49,6 +49,15 @@
 
 /* SourceSets index styles */
 
+.all-sets h1 a:link,
+.all-sets h1 a:hover {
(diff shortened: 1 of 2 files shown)


AudreyAltman:
@markbreedlove as per our conversation, I changed the link color to blue so it matches headers elsewhere on the site.  Now the only indication that the head is a link is that the cursor changes when you hover over it.  It's subtle, but it maintains aesthetic with other headers on the site.

AudreyAltman:
Also, I tested this in my browser and it works as expected.

PR #108, opened 2016-01-26
Add permissions system

markbreedlove:
Add a permissions system, with administrator roles.

See https://issues.dp.la/issues/8163

The specification was to add a class of view-only users (who can view unpublished sets, but not modify or create them), and a permissions system is necessary to achieve this.

### DIFF
--- a/app/views/registrations/edit.html.erb
+++ b/app/views/registrations/edit.html.erb
@@ -14,15 +14,55 @@
              url: registration_path(resource),
              html: { class: 'pss-admin-form' }) do |f| %>
 
-  <%= f.label :email %><br />
-  <%= f.email_field :email, autofocus: true %>
+  <fieldset>
+    <%= f.label :email %><br />
+    <%= f.email_field :email, autofocus: true %>
+  </fieldset>
 
-  <%= f.label :password %>
-  <%= f.password_field :password, autocomplete: "off" %>
-  <div><i>(leave blank if you don't want to change it)</i></div>
+  <% if current_admin.id == resource.id %>
+  <fieldset>
+    <%= f.label :password %>
+    <%= f.password_field :password, autocomplete: "off" %>
+    <div class="pss-admin-form-comment">
+      <i>(leave blank if you don't want to change it)</i>
+    </div>
+    <%= f.label :password_confirmation %>
+    <%= f.password_field :password_confirmation, autocomplete: "off" %>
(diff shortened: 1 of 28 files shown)


app/views/registrations/edit.html.erb
```diff
@@ -14,15 +14,43 @@
...
+    <%= f.radio_button :status, :reviewer,
+                       checked: resource.status == 'reviewer' %>
+    <%= f.label :status_reviewer, "reviewer" %>
+    <%= f.radio_button :status, :editor,
+                       checked: resource.status == 'editor' %>
+    <%= f.label :status_editor, "editor" %>
+    <%= f.radio_button :status, :manager,
+                       checked: resource.status == 'manager' %>
+    <%= f.label :status_manager, "manager" %>
+  </fieldset>
```
AudreyAltman: Could we add some brief text explaining the difference permissions afforded to reviewer, editor, and manager?

markbreedlove:
@AudreyAltman: I've added some explanatory text that might clarify the roles.

AudreyAltman:
When testing this in my local browser, I'm finding that I can't change an admin's role/status.    If I am logged in as a manager, access an `admins#registrations#edit` form for myself or any other admin, and try to change the role, it doesn't persist.

markbreedlove:
OK, I'll look into that!

markbreedlove:
@AudreyAltman that last commit oughta fix it.

AudreyAltman:
@markbreedlove yep, that fixed it!  

I think it would be good on the source_set, set, and guide views to hide links to functionality that isn't available to reviewers, but that seems out-of-scope for this ticket - I will make a new Redmine ticket to follow up on it.

All in all, this looks great.  I think it's ready to go.  Do you want to :squash: before I :merge:?

markbreedlove:
Oh yeah, I guess I overlooked some places where there are create-content links. Do the main navigational links at least get removed for you if you're a reviewer? (The links to "Administrators" and so forth?)

markbreedlove:
I will squash it if all's well with the nav links.

AudreyAltman:
@markbreedlove yes, the navigational links accurately reflect what actions are available to all three roles.

markbreedlove:
OK, @AudreyAltman, should be good to go.

AudreyAltman:
:rocket:

PR #112, opened 2016-02-24
carousel showing more sources on source#show view

AudreyAltman:
This introduces a carousel on the `source#show` view that links to other sources from the same set.

The carousel is created using the [slick](https://github.com/kenwheeler/slick) JavaScript library.  The carousel is responsive, displaying a different number of sources depending on screen size.

The `source` model has an instance method, `related_sources`, that returns all of the _other_ sources from this source's associated set.  The `source` model also has a default sort order (before, the sort order was defined in the view, which could lead to inconsistencies in the application).

A new partial, `_source_tile.html.erb`, contains the HTML representation of a source "tile" (a stylized thumbnail and title link to the source), used on both the `source#show` and `source#index` pages.  Some tweaks to the CSS ensure that it is usable in both places.

If JavaScript is disabled in the browser, the user instead sees a hyperlinked list of sources.  A CSS class, `js-off`, can be applied to HTML elements that should be hidden if JavaScript is disabled.  If JavaScript is enabled, a function removes the `js-off` class from all HTML elements when the page loads.

This has been deployed to staging for testing purposes, and approved by Franky.

This addresses [ticket #7924](https://issues.dp.la/issues/7934).

### DIFF
--- a/app/assets/javascripts/carousel.js
+++ b/app/assets/javascripts/carousel.js
@@ -1,0 +1,32 @@
+$(function() {
+  $(document).ready(function() {
+
+    /* 
+     * Initiate slick slider carousel. Slick is an external javascript library.
+     * @see config/initializers/assets.rb
+     * https://github.com/kenwheeler/slick
+     */
+    $('.carousel').slick({
+      infinite: true,
+      arrows: true,
+      slidesToShow: 3,
+      slidesToScroll: 3,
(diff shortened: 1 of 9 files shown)


anarchivist:
:+1: but needs to be rebased.

AudreyAltman:
Rebased to develop.

PR #114, opened 2016-02-25
hide resort button when filter selected

AudreyAltman:
This corrects a mistake in the `source_set#index` view.  The mistake occurred with a filter tag was included in the URL and the browser's JavaScript was disabled.  Before the, "Re-sort" continued to be visible even though the drop-down menu was hidden (see screenshot below).  This was confusing, and inconsistent with parallel behavior when JavaScript is enabled.  Now, the "Re-sort" button only shows if the dropdown menu is also visible.

screen shot 2016-02-22 at 10 04 02 am

I tested this in my local browser and it is working as expected.

This addresses [task 8286](https://issues.dp.la/issues/8286).

### DIFF
--- a/app/views/source_sets/index.html.erb
+++ b/app/views/source_sets/index.html.erb
@@ -43,6 +43,9 @@
           <%= select_tag('tags[]',
                        options_for_select(tag_filter_options(tags),
                                           selected_slugs_in(tags))) %>
+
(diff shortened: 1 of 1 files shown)


anarchivist:
:cactus: Rebasez-ci, s'il vous plaît! :fr:

AudreyAltman:
Rebased to develop and ready to merge :cactus:

PR #117, opened 2016-02-29
Fix/mobile layout margins

AudreyAltman:
This adds an outermost margin for the DOM element `div.primary-source-sets`, which contains main content (not header or footer).  It also moves one DOM element outside of `div.primary-source-sets` b/c it should not have an external margin and does not contain content specific to primary source sets.  Finally, it tweaks some paddings and widths for other DOM elements so that the layout overall looks good with the the new outermost margin.  It has been tested on android phone and iphone.  

This addresses [ticket #8232](https://issues.dp.la/issues/8232).  It has been tested and approved by Franky on staging.

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -400,4 +400,35 @@ input.form-submit {
 
 .carousel-container .source-container {
   margin: 16px;
+
+/* Mobile layout */
+
+@media only screen and (max-width: 680px) and (min-width: 0px) {
+  .primary-source-sets {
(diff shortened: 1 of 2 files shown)


anarchivist:
Looks good; needs rebasing. Having screenshots to indicate the visual difference for cases like this might be useful for reviewers - I don't think this is something to resolve immediately, but perhaps we could look at using something like [Wraith](https://github.com/BBC-News/wraith) to generate these for us.

AudreyAltman:
Good suggestion about screenshots!  Will look into Wraith for the future.  Rebased to develop.

PR #118, opened 2016-03-02
views for reviewers

AudreyAltman:
This introduces views specific to reviewers.  Reviewers are a class of admins who need to see unpublished sets (and their associated sources and guides), but do not need access to edit/delete functions. This filters out links to these pages on set, source, and guide views for reviewers.

Note that reviewers already cannot access edit or delete routes, and if they try they will be redirected to more benign pages. But to avoid confusion, it would be best if link to edit/delete pages didn't appear to reviewers at all.  

There are also a couple small changes to the views to clarify information for reviewers, who do not have intimate knowledge of current workflows.

Finally, this adds tests for manager, editor, and reviewer admins for the associated views.

This addresses [ticket #8263](https://issues.dp.la/issues/8263).  It has been approved by Franky on staging.

### DIFF
--- a/app/views/guides/show.html.erb
+++ b/app/views/guides/show.html.erb
@@ -103,11 +103,20 @@
 <% if admin_signed_in? %>
   <h2>Admin info</h2>
 
-  <p>
-    <%= link_to "Edit", edit_guide_path(@guide) %>
-    |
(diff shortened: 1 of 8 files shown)


anarchivist:
:rocket:

PR #119, opened 2016-03-03
Add URI validation and helper text to tag & vocab forms

AudreyAltman:
This adds helpful text to `tag` and `vocab` forms explaining what tags, vocabularies, and tag URIs are.  It also changes the expected data type for the form form `text` to `uri`.  This will perform a build-in validation to make sure the the user-submitted data is a properly-structured URI.  If the user does not enter a URI, the interface shows a tooltip saying they need to enter a URI (screenshot below).

screen shot 2016-03-03 at 4 42 36 pm

This addresses [ticket #8229](https://issues.dp.la/issues/8229).

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -3,6 +3,7 @@ class Tag < ActiveRecord::Base
   has_and_belongs_to_many :source_sets
   has_and_belongs_to_many :vocabularies
   validates :label, presence: true, uniqueness: true
+  validates :uri, format: { with: URI.regexp }, if: proc { |a| a.uri.present? }
 
   ##
   # FriendlyId generates a human-readable slug to be used in the URL, in place
--- a/app/views/tags/_form.html.erb
+++ b/app/views/tags/_form.html.erb
@@ -24,7 +24,8 @@
   </p>
 
   <p>
-    <strong><%= f.label :uri %></strong>
+    <strong><%= f.label(:uri, 'URI') %></strong>
+    <em>If this tag correlates to a term from a controlled vocabulary with a permanent URI, you can add it here.</em><br/>
     <%= f.text_field :uri %>
   </p>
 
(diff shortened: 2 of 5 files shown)


anarchivist:
Is there a possibility of changing the validation message to read `Please enter a URI.` instead of `Please enter a URL.`?

AudreyAltman:
You can change the validation message, but it's a client-side HTML5 validation, so you have to override the browser behavior with JavaScript.  It's a bit fussy (example [here](http://jsfiddle.net/robertc/X8UXQ/190/)) so I'm on the fence about whether or not it's worth it.  What do you think, @anarchivist ?

anarchivist:
How is this handled by Rails? I guess I don't have a strong opinion about it, but it would be nice if the language were consistent.

AudreyAltman:
@anarchivist the Rails `url_field` helper method creates an HTML `input` tag with `type=url`, which triggers a client-side validation upon submission.  See http://apidock.com/rails/ActionView/Helpers/FormHelper/url_field

anarchivist:
OK, I think this is fine then. I'd mistakenly thought you could override the validation message using the `:options` parameter for `#url_field`.

AudreyAltman:
Upon further investigation, I realized that HTML5 validation is not supported by Safari, so I am going to find a different solution.

AudreyAltman:
@anarchivist I changed the validation so that it happens in the model.  This is safer, b/c we don't have to depend on the browser to run the validation.  It also allows us to control the message that is shown to the user if they enter an incorrect validation.  Now, if the user enters an incorrectly-formatted URI, this message will appear:

screen shot 2016-03-07 at 2 56 00 pm

anarchivist:
OK, looks good. Needs to be rebased. Thanks!

AudreyAltman:
@anarchivist rebased and ready merge

PR #120, opened 2016-03-04
back link and link to guide on source#show

AudreyAltman:
This introduces a back link on the `source#show` page.  If the user come to a `source#show` page from a set or guide, the back link directs them back to the set or guide.  

We expect that users looking through a set or reading through a guide will need to do a lot of clicking back and forth between the set or guide and the sources contained within them.  The back links should facilitate this type of user experience. The back links are also operationally and stylistically akin to those on DPLA item pages.

I had initially considered including `back_uri`as a parameter of the `source#show` URL, a la:
 
    http://dp.la/item/eed4628bb2544f92665632e85d350963?back_uri=http%3A%2F%2Fdp.la%2Fsearch%3Futf8%3D%25E2%259C%2593%26q%3Dhamster

However, links _to_ `source#show` _from_ guides (and possibly sets) are included in user-generated text blocks.  See for examples the links on [this guide](http://dp.la/primary-source-sets/guides/teaching-guide-exploring-the-beginnings-of-rock-n-roll) under the "Discussion questions" section.  This text (including the links) were written by an admin and submitted via a form.  Therefore, we cannot count on these links including a `back_uri` parameter.  

In order to create a reliable, consistent fulfillment of the requirement, this instead parses the referrer data in the HTTP request.  If the referrer is a set or guide, the back link is rendered.

This also includes a permanent link to any associated teaching guides on the `source#show` view, as per feedback from Franky when the feature was tested on staging.

This addresses [ticket #8269](https://issues.dp.la/issues/8269).

### DIFF
--- a/app/helpers/sources_helper.rb
+++ b/app/helpers/sources_helper.rb
@@ -13,4 +13,38 @@ module SourcesHelper
       render partial: 'video'
     end
   end
+
+  ##
+  # Render a back link to the referring page if the referring page is a guide or
+  # set.  Links from guides or sets to sources may be created by admins, so the
+  # applicaiton cannot expect that back URI's can be included in params.
+  # Therefore, it is necessary to parse the referring URI to see if it is a
+  # guide or set, and render a back link accordingly.
+  #
+  # @return [String] HTML back link
+  def render_back_link
+    return unless request.referer.present?
+    uri = URI(request.referer)
+    # return if referrer is an offsite link
+    return unless uri.host == request.host
+
+    uri_path = uri.path
+    # return if referrer does not have relative url root
+    return unless uri_path.start_with?(Settings.relative_url_root)
+
+    uri_path.slice!(Settings.relative_url_root)
+    path = Rails.application.routes.recognize_path(uri_path) rescue nil
+
+    return unless path.present?
+    return unless path[:controller] == 'source_sets' || 'guides'
+    return unless path[:action] == 'show'
+
+    link_to "« back to #{path_label[path[:controller]]}", request.referer
+  end
+
+  private
+
+  def path_label
+    { 'source_sets' => 'set', 'guides' => 'guide' }
+  end
 end
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -315,6 +315,11 @@ audio {
   margin: 2em 0;
 }
 
+.related-guides {
+  margin-top: 3em;
+  margin-bottom: 3em;
+}
+
 /* Admin interfaces */
 
 input.form-submit {
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -1,5 +1,6 @@
 class Source < ActiveRecord::Base
   belongs_to :source_set
+  has_many :guides, through: :source_set
   has_many :attachments, dependent: :destroy
 
   has_many :videos, through: :attachments,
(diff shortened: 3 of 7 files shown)


app/helpers/sources_helper.rb
```diff
@@ -13,4 +13,35 @@ def render_media_asset
...
+  # applicaiton cannot expect that back URI's can be included in params.
+  # Therefore, it is necessary to parse the referring URI to see if it is a
+  # guide or set, and render a back link accordingly.
+  #
+  # @return [String] HTML back link
+  def render_back_link
+    return unless request.referer.present?
+    uri = URI(request.referer)
+    # return if referrer is an offsite link
+    return unless uri.host == request.host
```
markbreedlove: Is this going to return with no action the way we want it to if there's a link from somewhere else on the same domain, e.g. the main DPLA site linking to a source page?

AudreyAltman: Good question. I [amended the code](https://github.com/dpla/primary-source-sets/pull/120/files#diff-89d53548c3df2e0e1fd19f6e08527037R33) to return with no action if the URL does not have the correct relative root (ie. `/primary-source-sets`).  I also added a [test](https://github.com/dpla/primary-source-sets/pull/120/files#diff-08b75f9ca481ff07212281153c81226cR57) for scenario you described above.

app/helpers/sources_helper.rb
```diff
@@ -13,4 +13,36 @@ def render_media_asset
...
+
+    uri_path = uri.path
+    # return if referrer does not have relative url root
+    return unless uri_path.slice!(Settings.relative_url_root).present?
```
markbreedlove: I was curious about the use of `slice!` because I've always used `slice` functions in other languages / situations for extracting ranges of things, not string comparisons. When I ran something similar through `irb`, I got this error:
```
irb(main):020:0> '/test'.slice!('/test').present?
NoMethodError: undefined method `present?' for "/test":String
	from (irb):20
	from /Users/mark/.rbenv/versions/1.9.3-p547/bin/irb:12:in `<main>'
```

Do you know if Rails is overriding `String` in some way to add `#present?`?

I'm not sure if this is a big issue, but it wasn't clear reading it the first time what was being compared.  Is there a reason why `#start_with?` isn't used here?

markbreedlove: You'll notice I accidentally used ruby 1.9.3, but I get the same error in 2.1.3.

AudreyAltman: `present?` is a, as you suspected, Rails method - here's the documentation: http://apidock.com/rails/v3.1.0/Object/present%3F

I didn't use `#start_with` just b/c I was trying to be efficient - I have to do the `slice!` anyway b/c I want to remove the relative root from `uri_path` before I do `Rails.application.routes.recognize_path(uri_path)` in the next line.  But, I get what you're saying that it's hard to read and understand what's going on.  Do you think it would be better if I did:

    return unless uri_path.start_with?(Settings.relative_url_root)
    uri_path.slice!(Settings.relative_url_root)

markbreedlove: Sure, I think that would be more clear.

AudreyAltman:
Not sure why the jasmine tests are suddenly failing - the same thing is happening on [PR#124](https://github.com/dpla/primary-source-sets/pull/124).  They are passing locally.  Will investigate this afternoon.

AudreyAltman:
Okay @markbreedlove I [refactored](https://github.com/dpla/primary-source-sets/pull/120/files#diff-89d53548c3df2e0e1fd19f6e08527037R32) that bit of confusing code for clarity, and the tests are passing.

markbreedlove:
Thanks, @AudreyAltman, I'll merge this now, if that's still OK.

AudreyAltman:
@markbreedlove yes, it's okay with me to merge.  Thanks for your feedback!

PR #122, opened 2016-03-04
order sets by published date

AudreyAltman:
This replaces the ability to sort sets by their created date with the ability to sort sets by their published date.  This will make management of sets easier for admins.

* A `datetime` column, `published_at`, is added to the `source_sets` table.
* The value for the `published_at` column is set to the current value of the `created_at` column _if sets are already published_.  This will preserve the current sort order in production.  We do not currently have rspec tests for migrations (maybe something to look into for the future), but I tested this locally to ensure that it works as expected.
* A new `before_save` method is added to the `SourceSet` model to handle persistence of the publication timestamp.  If the set is being published upon save, the publication timestamp is set to the current `datetime`.  If a previously-published set is being otherwise edited, the publication timestamp is _not_ changed.  If a set is being un-published, the publication timestamp is cleared.
* The `order_by` method in `SourceSet` is changed to sort by `published_at` instead of `created_at`.

This addresses [ticket #8236](https://issues.dp.la/issues/8236).

### DIFF
--- a/app/models/source_set.rb
+++ b/app/models/source_set.rb
@@ -9,6 +9,7 @@ class SourceSet < ActiveRecord::Base
   validates :name, presence: true
   validates_numericality_of :year, only_integer: true, allow_nil: true,
                                    less_than_or_equal_to: Date.today.year
+  before_save :check_publish_date
 
   ##
   # FriendlyId generates a human-readable slug to be used in the URL, in place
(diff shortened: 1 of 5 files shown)


app/models/source_set.rb
```diff
@@ -64,4 +65,18 @@ def self.with_tag(tag)
...
   end
   private_class_method :with_tag
+
+  private
+
+  ##
+  # If the set is being published, save the current timestamp.
+  # If the set was already published, do not save the current timestamp.
+  # If the set is unpublished, clear the timestamp.
+  def published_set
```
markbreedlove: Could this method name incorporate a verb, like "check publish date" or something similar, since it's not returning a published set, it's doing something to change the set?

markbreedlove:
Thanks, @AudreyAltman. My only comment is on the method name `published_set`. I'm happy to merge this pending your response to that.

app/models/source_set.rb
AudreyAltman: Great suggestion!  I'll make that change.

AudreyAltman:
@markbreedlove the method name `published_set` has been changed to `check_publish_date` as per your suggestion.  Should be ready to :merge:.

PR #124, opened 2016-03-07
posters for sets & guides

AudreyAltman:
This introduces a new feature, "posters", which are static image representations of sets and guides to be used for marketing.  Static images include names of sets or guides, their associated images, and the DPLA logo.  Specifically, these images will be used when submitting sets and guides to PBS learning media.  Admins will screenshot the images for local use.  While this is admittedly not ideal, we are saving time by not trying to implement HTML to image file functionality.

There are new routes, controllers, and views, but no new models as all the data needed to construct the posters are in `SourceSet` and `Guide`.  JavaScript is used for dynamic styling.  Posters are only visible to admins.

A sample poster:

screen shot 2016-03-07 at 4 32 01 pm

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,9 @@ gem 'zencoder', '~>2.5'
 gem 'navigasmic', '~>1.0'
 gem 'turnout', '~>2.0'
 gem 'access-granted', '~> 1.0.0'
+# As of jasmine-jquery-rails v2.0.3, jasmine-jquery-rails is not compatible with
+# rake 11.0.0
+gem 'rake', '< 11.0'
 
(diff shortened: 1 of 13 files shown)


no-reply:
Awesome. :rocket: 

Nice, clean, well tested code. :100:

Looks like it needs a rebase on `develop`. Happy to merge when that's up.

AudreyAltman:
Not sure why the jasmine tests are failing after rebase - I'll look into this afternoon.

AudreyAltman:
@no-reply okay, tests are passing now.

no-reply:
Yay!

PR #125, opened 2016-03-14
Allow admins to order tags within vocabularies

AudreyAltman:
This allows admins to order tags within vocabularies.  This is important for vocabularies such as "time period", where admins will want tags to display to users in chronological order.  Currently, tags are ordered by their created date, so admins have to consciously create tags in the order in which they want them to display. 

Tags and vocabularies have a many-to-many relationship.  This changes the name of their join table from `tags_vocabularies` to `sequences` for clarity.  A new field, `position`, has been added to the `sequences` table to record the sequential order in which tags should appear within vocabularies.  A final migration sets a `position` value for all existing rows in the join table.

A new model, `Sequence` has methods that check/amend `position` values when a sequence is saved or destroyed.  The `Tag` and `Vocabulary` models are edited to assert their relationship with sequence.  Since the name of the table is no longer `tags_vocabularies`, some things need to be made explicit that were previously taken care of through naming conventions (ie. "Rails magic").

A new route is added to handle re-sorting sequences.  This route sends a `POST` request to update the `sequences` table.  This route is only available to admins (managers and editors).

Resorting sequences in the UI happens on the `vocabularies#show` view.  A JavaScript UI plugin introduces drag-and-drop sorting.  While this adds an additional dependency, it makes for good user experience, and also helps to ensure that the items within a sequence are correctly incremented (as opposed to human-entered values, which could be incorrectly incremented).

This addresses [ticket #8243](https://issues.dp.la/issues/8243).  It has been tested locally and on staging.

### DIFF
--- a/app/assets/javascripts/poster.js
+++ b/app/assets/javascripts/poster.js
@@ -17,8 +17,6 @@ $(function() {
         height += $(this).height();
       });
 
-      console.log(height);
-
       /*
        * If the height of all the lefthand images isn't enough to fill the
        * space, repeat the the images.
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem 'coffee-rails', '~> 4.0.0'
 gem 'therubyracer',  '~> 0.12.2', platforms: :ruby
 gem 'sass-rails', '~> 4.0.3'
 gem 'jquery-rails', '~> 3.1.3'
+gem 'jquery-ui-rails', '~> 5.0'
 gem 'config', '~> 1.0.0'
 gem 'redcarpet', '~> 3.3'
 gem 'friendly_id', '~> 5.1.0'
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -11,5 +11,6 @@
 // about supported directives.
 //
 //= require jquery
+//= require jquery-ui/sortable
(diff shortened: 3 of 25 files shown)


app/assets/javascripts/poster.js
```diff
@@ -17,8 +17,6 @@ $(function() {
         height += $(this).height();
       });
 
-      console.log(height);
-
```
AudreyAltman: This was something I forgot to delete from a previous PR.

markbreedlove:
Thanks, @AudreyAltman, this is great. The only thing that threw me was the term "sequence" which I could imagine applying to various things. Does it break anything to call it a tag sequence? Someone has to have the experience to know that a sequence, or tag sequence, is something that pertains to vocabularies, or has to have read the doc comment in the Sequence model.

markbreedlove:
I tested the app in my browser, btw, and it worked fine.

AudreyAltman:
@markbreedlove I agree, I think tag sequence is a much better name!  I'll make that change.

markbreedlove:
Discussed elsewhere:  We'll go ahead and merge this as-is, and there can be followup work to fine-tune the name of the controller and model. There's a deployment issue that prevents this from being an easy change at the moment.

AudreyAltman:
rebased and ready to merge

PR #126, opened 2016-03-15
Add link to provider's digital resource.

markbreedlove:
This addresses https://issues.dp.la/issues/8098, adding a link to the provider's digital resource on the source page.

I didn't add any tests, because the real work is handled by DPLibrary, which has -- or should have -- its own tests. Let me know if you think the `logger.error` call in `ApiQueryer` should be tested.

Is it best to have the API functionality in a concern?  I think it probably is, but want to make sure that sits well with you.

I have `ApiQueryer#dpla_items` taking variable arguments, because originally I thought it would be like a model's `find` method, which does just that.  I'm not so sure now, though. Let me know if you think it would be best simply taking an array. I don't really know the circumstances under which we'll be querying multiple item IDs some day.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,8 @@ gem 'access-granted', '~> 1.0.0'
 # As of jasmine-jquery-rails v2.0.3, jasmine-jquery-rails is not compatible with
 # rake 11.0.0
 gem 'rake', '< 11.0'
+gem 'DPLibrary', git: 'https://github.com/dpla/DPLibrary.git',
+                 branch: 'feature/add-new-prop-and-id-path'
 
 group :test, :development do
   gem 'rspec-core', '~> 3.3.2'
--- a/app/controllers/concerns/api_queryer.rb
+++ b/app/controllers/concerns/api_queryer.rb
@@ -1,0 +1,25 @@
+##
+# ApiQueryer: a mixin for controllers that need to query the DPLA API
+# and receive a DPLibrary::DocumentCollection
+#
+module ApiQueryer
+  extend ActiveSupport::Concern
+
+  ##
+  # Given one or more DPLA item IDs, return an array of items
+  #
+  # @param [Array<String>] *ids  One or more item IDs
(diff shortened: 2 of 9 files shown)


AudreyAltman:
@markbreedlove I think this is great!  I like that API functionality is in a controller concern, that seems like a natural fit.  I also think it's quite nice that `ApiQueryer#dpla_items` takes variable arguments.  I don't anticipate any reason why that would be a problem, and it's nice that it mirrors the behavior of `find`.  

I don't think you need to write a test for the `logger.error` call.  If you wanted, you could write a test to check that DPLibrary is called.  The syntax here isn't exact, but something like:

    expect(DPLibrary::DocumentCollection).to receive(:new).with(apis)

But if that feels too fussy, you can leave it out.

Thanks for all you work on this.  I'm so excited to get the API functionality integrated into the application!

markbreedlove:
Thanks for your quick feedback, @AudreyAltman!  After submitting this, I've realized that maybe I should set the API key in an initializer.  I'm going to see how that works out.

markbreedlove:
@AudreyAltman, I've added some tests to verify that the controller is calling the concern method, the concern method is querying the API, and bad queries will be logged.
I should squash this before it's merged.

AudreyAltman:
@markbreedlove looks good!  Let me know when it's squashed and rebased, and I'll merge it. Thanks!

markbreedlove:
Actually ... yet another thing. I meant to add a view test, but forgot ... coming up ...

markbreedlove:
@AudreyAltman: there's just one more commit -- I think my last one -- to add some tests for the view, and for the return value of `ApiQueryer#dpla_items` when there's a failure.  I'll squash this again if it's OK.

AudreyAltman:
:+1: definitely worthwhile to add those view tests :)

markbreedlove:
@AudreyAltman: It's squashed now, and ready to merge when CI passes again.

AudreyAltman:
:fireworks:

PR #127, opened 2016-03-15
add default thumbnails

AudreyAltman:
This adds default thumbnails for sources.  If a thumbnail is not present for a source, a default image is shown instead.  This will be especially helpful for audio and video sources - admins generally do not make custom thumbnails for these, and can rely on defaults.

This addresses [ticket #8064](https://issues.dp.la/issues/8064).


markbreedlove:
That looks good, and I can merge it when it's rebased.

AudreyAltman:
@markbreedlove rebased - should be ready to merge when tests pass

PR #129, opened 2016-03-16
change header text for related sets and sources

AudreyAltman:
This changes the `<h2>` text for related set and sources for clarity.

It also sets a `@source_set` variable for `source#show` - I noticed there were several calls to `@source.source_set`, and wanted to reduce calls to the database.

This addresses [ticket #8323](https://issues.dp.la/issues/8323).


markbreedlove:
I'll merge after it's rebased!

PR #131, opened 2016-03-18
move javascript to head with defer attribute

AudreyAltman:
This moves all included javascript files to the HTML `<head>` and uses the `defer` tag to ensure that the javascripts do not load until after the page has finished parsing.  Before, we were inconsistently using the `defer` tag and putting included javascript files in both the `<head>` and after the `<body>`.

The `defer` attribute is supported by all major browsers of reasonably recent versions: http://www.w3schools.com/tags/att_script_defer.asp

This has been tested in a local deployment, but since we do not have full tests for all javascript functionality, it should be deployed and manually tested by a second set of eyes.

This addresses [ticket #8248](https://issues.dp.la/issues/8248).

### DIFF
--- a/app/views/sources/show.html.erb
+++ b/app/views/sources/show.html.erb
@@ -8,13 +8,13 @@
                                           trailing_slash: true) %>' />
   <link rel="stylesheet" type="text/css" href= 'http://cdn.jsdelivr.net/jquery.slick/1.5.9/slick.css' />
   <link rel="stylesheet" type="text/css" href= 'http://cdn.jsdelivr.net/jquery.slick/1.5.9/slick-theme.css' />
+  <script type="text/javascript"
+          src="http://cdn.jsdelivr.net/jquery.slick/1.5.9/slick.min.js"
+          defer></script>
+  <%= javascript_include_tag 'carousel', defer: 'defer' %>
 <% end %>
 
 <% content_for :foot_script do %>
-  <%= javascript_include_tag 'lightbox' %>
-  <%= javascript_include_tag 'carousel' %>
-  <script type="text/javascript"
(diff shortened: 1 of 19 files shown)


app/views/sources/show.html.erb
```diff
@@ -8,13 +8,13 @@
...
   <link rel="stylesheet" type="text/css" href= 'http://cdn.jsdelivr.net/jquery.slick/1.5.9/slick.css' />
   <link rel="stylesheet" type="text/css" href= 'http://cdn.jsdelivr.net/jquery.slick/1.5.9/slick-theme.css' />
+  <script type="text/javascript"
+          src="http://cdn.jsdelivr.net/jquery.slick/1.5.9/slick.min.js"
+          defer></script>
+  <%= javascript_include_tag 'carousel', defer: 'defer' %>
 <% end %>
 
 <% content_for :foot_script do %>
-  <%= javascript_include_tag 'lightbox' %>
```
AudreyAltman: I deleted lightbox entirely because I don't think we are using it now that open seadragon has been implemented.  Is this correct?

markbreedlove: That sounds right, yes.

markbreedlove:
This seems to work well in a simulated production environment.

With all of the assets compiled, in production, `application*.js` comes first in the `<head>` without a `defer` attribute, and the other `<script>`s come after with `defer`. Is that intentional?

AudreyAltman:
@markbreedlove yes, that is intentional.  Things break if `application.js` is given a `defer` attribute.

markbreedlove:
This is all good. I can confirm also that lightbox is no longer used.
Thanks, @AudreyAltman!

PR #132, opened 2016-03-18
social media share buttons

AudreyAltman:
This adds social media share buttons to sets, sources, and guides (show views only).  The HTML and javascript are copied exactly from the frontend application for consistency and ease of maintenance.  A small amount of CSS improves the display.

This addresses [ticket #7922](https://issues.dp.la/issues/7922).

### DIFF
--- a/app/assets/javascripts/social.js
+++ b/app/assets/javascripts/social.js
@@ -1,0 +1,19 @@
+$(window).load(function() {
+
+  ///// Social media buttons
+  if ($('.shareSave').length) {
+    (function(d, s, id) {
+      var js, fjs = d.getElementsByTagName(s)[0];
+      if (d.getElementById(id)) return;
+      js = d.createElement(s); js.id = id;
+      js.src = "//connect.facebook.net/en_US/all.js#xfbml=1";
+      fjs.parentNode.insertBefore(js, fjs);
+    }(document, 'script', 'facebook-jssdk'));
+
+    (function() {
+      var po = document.createElement('script'); po.type = 'text/javascript'; po.async = true;
+      po.src = 'https://apis.google.com/js/plusone.js';
+      var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(po, s);
+    })();
+  }
+});
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -285,6 +285,8 @@
 
 .breadcrumbs {
   padding: 0 0 10px 0;
+  display: inline-block;
+  float: left;
 }
 
 iframe {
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/social.js
```diff
@@ -0,0 +1,19 @@
+$(window).load(function() {
+
+  ///// Social media buttons
+  if ($('.shareSave').length) {
+    (function(d, s, id) {
```
markbreedlove: I'm not sure that the closures here and on line 13 are necessary for privacy / variable scope reasons, considering that this is within an anonymous function given to `load()` above.  Do you have a position in favor of their existing here?

app/assets/javascripts/social.js
```diff
@@ -0,0 +1,19 @@
...
+      var js, fjs = d.getElementsByTagName(s)[0];
+      if (d.getElementById(id)) return;
+      js = d.createElement(s); js.id = id;
+      js.src = "//connect.facebook.net/en_US/all.js#xfbml=1";
+      fjs.parentNode.insertBefore(js, fjs);
+    }(document, 'script', 'facebook-jssdk'));
+
+    (function() {
+      var po = document.createElement('script'); po.type = 'text/javascript'; po.async = true;
+      po.src = 'https://apis.google.com/js/plusone.js';
```
markbreedlove: If the `js.src` declaration above on line 9 assigns a scheme-relative URL, could this do the same? (Removing `https:`)

app/assets/javascripts/social.js
```diff
@@ -0,0 +1,19 @@
...
+      var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(po, s);
```
markbreedlove: Finally, here and above on line 14, could these statements be made one per line?

AudreyAltman:
@markbreedlove I'm happy to improve the JavaScript as per your suggestions, but I just want to check about one thing - I copied this script exactly from the frontend app and didn't change it b/c it seemed like it might be difficult to maintain two versions of the same code.  Do you think it's better to improve the code, or to leave it so that it matches what's in the frontend?  Or maybe door number three is making the improvements in both places.

markbreedlove:
I vote for changing them here, and changing them later in the frontend when that's practical. The code isn't shared, so it's not synced anyway, and the maintinter of the code in either place doesn't know it's duplicated elsewhere, so one can't rely on it to be consistent. A further ticket might be to get rid of this problem.

Since these suggestions are related to readability and not functionality, my attitude is we might as well take the opportunity to improve the readability. Though I have to say I'm not super invested, considering it should be factored out into some common module down the road.

AudreyAltman:
Due to time constraints, @markbreedlove and I decided not to address his concerns about readability of the JavaScript function at this time.  They can be addressed later, and have been documented in this ticket: https://issues.dp.la/issues/8351

AudreyAltman:
@markbreedlove rebased

PR #133, opened 2016-03-25
validate that aggregation is only letters and numbers

AudreyAltman:
This adds a validation to the `Source` model ensuring that the value for an aggregation (ie. the ID of an item in DPLA) is comprised of letters and numbers.  This will help to prevent errors elsewhere in the system, where links are formed to items in the frontend using the aggregation (ie.  http://dp.la/item/90aade1f9d91965c921026d3b0dc2d8f). 

This addresses [ticket #8330](https://issues.dp.la/issues/8330).

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -34,7 +34,9 @@ class Source < ActiveRecord::Base
                         source: :asset,
                         source_type: 'Image'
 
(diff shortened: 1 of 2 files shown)


markbreedlove:
OK, looks good.

PR #136, opened 2016-04-01
Improve workflow links for building associations between objects.

AudreyAltman:
This improves the ways that admins can manage associations between objects using forms.

Before this PR, there were links on various #show views that allowed users to create _new_, associated objects.  For example, on the `/sets/123` view, there was a link that said "Add new tag", which linked to `/tags/new?source_set_id=123`.  The new tags form would be pre-populated with the set id to assert the association.  

Admins could also build associations between _existing_ objects on various #new and #edit forms.  For example, on `sets/123/edit`, there was a menu of existing tags that an admin could select in order to associate them with the set.

This was confusing for admins because they needed to go to separate views to do related tasks, ie. associating tags with a given set.

This improves the workflow by adding links on #show views to manage associations with _both new and existing_ objects.  Referring to the previous example, the `sets/123` view now has two links.  The first, "Create new tag", links to `/tags/new?source_set_id=123`.  The second, "add/remove tags", links to `sets/123/edit#tag`, on which they can select from a menu of existing tags.  The `#tag` jump link in the URI triggers the expansion of the tags menu, which would otherwise be collapsed by default.

This addresses [ticket #8355](https://issues.dp.la/issues/8355).

### DIFF
--- a/app/assets/javascripts/form.js
+++ b/app/assets/javascripts/form.js
@@ -29,6 +29,8 @@
     $('.expand-control').html('show').show();
     $('.expand-content').hide();
 
+    showExpansibleAnchors();
+
     /* 
      * Show/hide function for expansible content.
      * .expand-control is a DOM element that users can click to show or hide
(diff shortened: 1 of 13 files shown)


AudreyAltman:
Investigating failing tests...

AudreyAltman:
All tests passing.  Ready for review.

markbreedlove:
@AudreyAltman, I'm happy to merge this after it's rebased.

AudreyAltman:
@markbreedlove rebased!

PR #138, opened 2016-04-04
change static text in sidebar

AudreyAltman:
This changes the static HTML text on `/source/show` and the accompanying CSS.  This addresses [ticket #8356](https://issues.dp.la/issues/8356).


AudreyAltman:
Rebased to develop

PR #139, opened 2016-04-04
fix header layout on source page

AudreyAltman:
This fixes the style of the `<h1>` element on `/source/show`.  Before this fix, the new social media share button was causing it to float in an odd and unattractive way:

screen shot 2016-04-04 at 4 34 57 pm

Now, it floats in a lovely and sensible way:

screen shot 2016-04-04 at 5 09 41 pm

This addresses [ticket #8357](https://issues.dp.la/issues/8357).

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -262,6 +262,10 @@ table td {
 
 /* Source */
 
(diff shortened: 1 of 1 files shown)


no-reply:
:+1: Lovely and sensible is the correct way to float, IMO.

PR #140, opened 2016-04-06
track use of items for hubs

AudreyAltman:
This uses Google Analytics to track use of hubs' items.

When a `/source/show` page is loaded in a browser window, a message is sent to Google Analytics with the DPLA item ID, provider name, contributing institution, and title.  It matches the message sent to Google Analytics when an item is viewed in the DPLA portal.

Testing on staging, the message sent to Google Analytics is:

screen shot 2016-04-06 at 4 24 15 pm

If you want to test yourself on staging you can use the [Google analytics debugger Chrome plugin](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en).

This addresses [ticket #8065](https://issues.dp.la/issues/8065).

### DIFF
--- a/app/controllers/sources_controller.rb
+++ b/app/controllers/sources_controller.rb
@@ -26,8 +26,7 @@ class SourcesController < ApplicationController
     ma = @source.main_asset
     @file_base_or_name = nil
     dpla_item = dpla_items(@source.aggregation).first  # see ApiQueryer
-    @digital_resource_url = dpla_item.url unless dpla_item.nil?
-    @provider_name = dpla_item.provider.name unless dpla_item.nil?
+    parse_dpla_item(dpla_item)
 
     if ma.present?
       @file_base_or_name =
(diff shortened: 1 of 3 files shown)


app/controllers/sources_controller.rb
```diff
@@ -94,4 +93,15 @@ def source_params
...
+  def parse_dpla_item(dpla_item)
+    return unless dpla_item.present?
+    @digital_resource_url = dpla_item.try(:url)
+    @provider_name = dpla_item.try(:provider).try(:name)
+    data_provider = dpla_item.try(:source)
+    intermediate_provider = dpla_item.try(:intermediate_provider)
+    @contributing_institution = 
+      [data_provider, intermediate_provider].compact.join('; ')
+    @title = Array(dpla_item.try(:title)).flatten.first
+  end
```
AudreyAltman: This is getting to be a lot of logic for a controller.  I'm curious to know if you think it would be beneficial to instantiate `dpla_item` as an object, with methods to get the `digital_resource_url`, `provider_name`, etc.?  I am on the fence about it myself.  I don't think we will need to parse any more attributes than what you see here, at least in the foreseeable future, so I don't anticipate this `parse_dpla_item` method to get any bigger.

anarchivist: Let's make a new issue to refactor this. I think there's a lot we could do to make this somewhat cleaner but at first glance we may want to look at diverging from `DPLibrary`'s implementation so we can get what we need out of the object.

PR #141, opened 2016-04-07
only show published sets in related sets slider

AudreyAltman:
This fixes the `related_sets` method of the SourceSet model to only return published sets, as opposed to both published and unpublished sets.  The `related_sets` method is used to populate the "Related sets" carousel on the `sets/show` view.  Since unpublished sets are not accessible to the public, they should not be included in the slider.

This is a high-priority fix as it is currently holding up data entry for the final content release.

This has been tested locally.  It addresses [ticket #8362](https://issues.dp.la/issues/8362).

### DIFF
--- a/app/models/source_set.rb
+++ b/app/models/source_set.rb
@@ -34,13 +34,13 @@ class SourceSet < ActiveRecord::Base
   end
 
   ##
-  # Returns sets with at least two matching tags (not including self).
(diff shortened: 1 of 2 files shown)


anarchivist:
:rocket: Rebase?

AudreyAltman:
Rebased and ready to :merge:

PR #151, opened 2016-04-19
Use Settings.app_scheme instead of hardcoded protocol in helper test

AudreyAltman:
This fixes a test that was failing for me locally.  In a helper test, it uses the value from `Settings.app_scheme` instead of a hardcoded `http://` to construct a URL.  This is related to the switch to SSL.


AudreyAltman:
Rebased to develop

PR #154, opened 2016-04-26
css for dimensional manipulations source, source and set show views

AudreyAltman:
fix/css-dimensional-manipulation

This uses CSS instead of JavaScript for most dimensional manipulation.  This will improve user experience for people with JavaScript disabled in their browsers.  The changes in this PR also improves UX overall by making the views slightly more responsive for all browsers.

Before, JavaScript was used to ensure that certain boxes on a page were the same height within a responsive design.  This instead uses the CSS3 flexbox feature.

There is one dimensional manipulation that is still handled with JavaScript, but I think it is outside the scope of this PR because 1) the solution would be complex and therefore time consuming, 2) it requires a different kind of solution than the one employed in this PR, and 3) the degradation of UX for those without JavaScript is tiny and strictly aesthetic, so it is not as high a priority.  See: https://github.com/dpla/primary-source-sets/blob/846e17bce19bd6d7f5d0dc52d84ceedb60dc5101/app/assets/javascripts/style.js#L29  I have made a [new ticket](https://issues.dp.la/issues/8371)
 to capture the work that needs to be done to address this issue.

This has been tested on Chrome, Safari, Firefox, Microsoft Edge, Android, and Safari mobile.  It should be tested on Internet Explorer before merge (I'm working on getting access to IE).

This addresses [#8230](https://issues.dp.la/issues/8230).

### DIFF
--- a/app/assets/javascripts/style.js
+++ b/app/assets/javascripts/style.js
@@ -12,38 +12,13 @@
    });
 
   $(document).ready(function() {
-    setAsideHeight();
     setNameContainerHeight();
   });
 
   $(window).resize(function() {
-    setAsideHeight();
     setNameContainerHeight();
   })
 })();
-
-/*
- * Extend the height of the aside module to be at least as tall as the rendered
- * media asset.
- */
-function setAsideHeight() {
-  var height = $('.source aside .module').outerHeight();
-  var minHeight = $('.source .media-inner-container').height();
-  if (minHeight > height) {
-    $('.source aside .module').outerHeight(minHeight);
-  }
-
-  // resize header on set page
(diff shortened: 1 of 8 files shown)


markbreedlove:
This looks good to me. I'll merge it once it's been rebased.

AudreyAltman:
I rebased to develop.  After the rebase, I had to make a couple small HTML and CSS changes: https://github.com/dpla/primary-source-sets/pull/154/commits/295849a3b8e78b9cd31b93aa388d451e9c1a99c4

AudreyAltman:
Squashed.  Should be ready to merge.

PR #155, opened 2016-04-26
add error message if javascript disabled

AudreyAltman:
If a user does not have JavaScript enabled in their browser, they cannot load media assets. Audio and video files will not play; images will not load; and the document viewer does not load interactive features, such as scroll and zoom.  Instead of trying and failing to render the media objects, this displays an error message to users explaining that they need JavaScript. 

The `div` containing the media object is given the class `js-off`, which by default does not display.  A JavaScript function displays all objects with this class when the page loads.  See https://github.com/dpla/primary-source-sets/blob/develop/app/assets/javascripts/style.js#L10

The rendered message looks like this:

screen shot 2016-04-26 at 1 23 33 pm

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -18,6 +18,13 @@ table td {
   padding-bottom: 0.2em;
 }
 
+.error {
+  background-color: #EDEDED;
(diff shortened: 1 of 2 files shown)


moltude:
👍

PR #156, opened 2016-04-26
Fix/set tile render

AudreyAltman:
#This improves the rendering of set tiles in two ways:

1) It expands the width of the tiles so that they take up the entire available width of the page.
2) They eliminate the occasional white line that was appearing under set tiles if they were not exactly the expected pixel dimensions.

This has been tested locally.  It addresses [#8371](https://issues.dp.la/issues/8371).

Before:

screen shot 2016-04-27 at 5 25 46 pm

After: 

screen shot 2016-04-27 at 5 27 05 pm

### DIFF
--- a/app/assets/stylesheets/primary_source_sets.css
+++ b/app/assets/stylesheets/primary_source_sets.css
@@ -92,6 +92,14 @@ table td {
 
 /* Source set tile */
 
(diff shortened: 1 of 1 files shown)


moltude:
👍

PR #157, opened 2016-04-27
custom language for auto-generated twitter post

AudreyAltman:
This adds custom language for an auto-generate tweet, which includes the page URL, page title and "@dpla".  In order to easily include the page title in the tweet, page titles are now accessible with `content_for`.  The twitter account name, `dpla`, is already defined in the settings file through aws: https://github.com/dpla/aws/blob/0e10b825dbb3d66fb3930d583376ceddc5fda9a8/ansible/group_vars/production#L40

Now, for example, when you click on the Twitter share button from `/primary-source-sets/`, you are taken to a page where you see this:

screen shot 2016-04-27 at 5 32 21 pm

This has been tested locally.  It addresses [ticket #8360](https://issues.dp.la/issues/8360).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -54,4 +54,18 @@ module ApplicationHelper
       image_name
     end
   end
+
+  ##
+  # Generate a Twitter web intent link using the URL and title of the current
+  # page.
+  #
+  # @return String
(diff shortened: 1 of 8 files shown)


markbreedlove:
That looks great. Merging ...

PR #158, opened 2016-04-27
allow basic controller tests to run

AudreyAltman:
This addresses a "FIXME" note that was left in a controller spec.  Apparently a previous iteration of the code base prevented it from running a shared example test.  Now, it runs just fine.  The instance variable mentioned in the comment, `@file_base_or_name`, is by default set to `nil` in the controller `:show` method, which seems to eliminate the need for a mock.


moltude:
👍

PR #159, opened 2016-04-27
add test contexts with and without frontend assets gem

AudreyAltman:
This addresses a "TODO" in the application helper spec.  It adds a shared context that allows the spec to test the output of two methods both when the frontend assets gem is defined, and when it is not defined.  Before, it was only testing with the assumption that the gem was defined.

This addresses [ticket #8140](https://issues.dp.la/issues/8140).

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -1,10 +1,9 @@
 require 'rails_helper'
 
 describe ApplicationHelper, type: :helper do
-  before do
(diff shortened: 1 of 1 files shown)


markbreedlove:
Looks good to me. Does anyone else need to review it? I will merge it once it's rebased.

AudreyAltman:
Rebased!  I don't think anyone else needs to review it, so you can go ahead and merge once the tests pass.

PR #160, opened 2016-04-27
fix link to apps

AudreyAltman:
This fixes the link to the apps page in the footer so it has the correct URL.  This has been tested locally.  It addresses [ticket #8384](https://issues.dp.la/issues/8384).


moltude:
looks good

PR #161, opened 2016-04-27
add content jumplinks

AudreyAltman:
This adds jumplinks to public-facing pages that improve accessibility of the website.  

Prior to this PR, there were jumplinks to HTML elements with `id='content'` in our accessibility menu (see https://github.com/dpla/primary-source-sets/blob/develop/app/views/shared/_jump_links.html.erb#L5).  However, we were not actually implementing these links.  This PR adds HTML elements with `id='content'` to all public-facing pages.

Now, for example, you could use the following link to access the main content of the page:  `/primary-source-sets#content`

This has been tested locally.  It addresses [ticket #8385](https://issues.dp.la/issues/8385).

### DIFF
--- a/app/views/guides/show.html.erb
+++ b/app/views/guides/show.html.erb
@@ -19,7 +19,7 @@
   <%= render_breadcrumbs separator: ' / ' %>
 </div>
 
-<div class='guide'>
+<div class='guide' id='content'>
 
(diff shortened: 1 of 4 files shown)


moltude:
looks good to me. I'll update this branch and merge.

PR #163, opened 2016-05-06
embed json-ld in html pages

AudreyAltman:
This embeds the JSON-LD markup for sets, guides, and sources within the HTML.  It is intended to boost SEO.  The JSON-LD partials already exist.

This addresses [#8187](https://issues.dp.la/issues/8187).


moltude:
this is 👍

PR #164, opened 2016-05-06
Feature/dpla item model

AudreyAltman:
This refactors the code that deals with getting and parsing data from DPLibrary.  Before, the call the DPLibrary was handled in a controller concern, and the data was parsed in the controller.  As we’ve added features that use more data from DPLibrary, this code got too bulky and complex for the controller.  This PR introduces a model, `DplaItem`, that handles the request to DPLibrary and parsing the response.  This change simplifies the role of the controller, and also gives views a more straightforward way of checking for a valid response from DPLibrary.

This PR also introduces a small change to the public-facing `source#show` view.  If DPLibrary does not return a valid response, it displays a message saying that the item is not currently available in the DPLA library.

A couple things to note 
* The `DplaItem` model is an `ActiveModel`b/c it does not interact directly with the database, but rather populates its objects with data from an external API (I used a similar pattern in `Krikri`). 
* Specs use doubles instead of factories for objects having to do with DPLibrary.  The way that DPLibrary initializes objects makes the use of factories cumbersome.

This addresses tickets [#8252](https://issues.dp.la/issues/8252) and [#8364](https://issues.dp.la/issues/8364).


markbreedlove:
OK, I think that looks good. The controller's needs have grown since March. Let me know if anyone else will be reviewing it, otherwise, I'll be happy to do the merge.

AudreyAltman:
Nope, you're the only one who needs to review.  Merge away!  And thanks for your feedback :)

PR #165, opened 2016-05-09
Feature/optimize db calls

AudreyAltman:
This optimizes calls to the database by refactoring to reduce the total number of database calls, and eager loading objects with `includes`.  I hope that it will reduce page load time.  

This has been tested locally.  It addresses [#8259](https://issues.dp.la/issues/8259).

### DIFF
--- a/app/controllers/source_sets_controller.rb
+++ b/app/controllers/source_sets_controller.rb
@@ -11,8 +11,10 @@ class SourceSetsController < ApplicationController
   def index
     @tags = get_tags_from_params
     @order = params[:order]
-    @published_sets = SourceSet.published.order_by(@order).with_tags(@tags)
-    @unpublished_sets = SourceSet.unpublished.order_by(@order).with_tags(@tags)
+    @published_sets = SourceSet.includes(:small_images).published
+                               .order_by(@order).with_tags(@tags)
+    @unpublished_sets = SourceSet.includes(:small_images).unpublished
+                                 .order_by(@order).with_tags(@tags)
   end
 
   def show
--- a/app/helpers/source_sets_helper.rb
+++ b/app/helpers/source_sets_helper.rb
@@ -24,11 +24,14 @@ module SourceSetsHelper
     return {} unless source_sets.present?
 
     Vocabulary.filterable.each_with_object({}) do |vocab, hash|
+      ss_ids = source_sets.map { |set| set.id }
+
       # Get all tags associated with the vocab.
       # Ignore tags that aren't associated with any of the source_sets.
-      tag_list  = vocab.tags.reject do |tag|
-        tags_for_sets(source_sets).exclude?(tag)
-      end
+      tag_list = Tag.joins(:source_sets, :vocabularies)
+                    .where(source_sets: { id: ss_ids })
+                    .where(vocabularies: { id: vocab.id })
+                    .uniq
 
       # Return the vocab and associated tag list.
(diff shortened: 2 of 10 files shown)


moltude:
@AudreyAltman  I've done some reading on includes/joins and I think this looks good. I haven't done any performance testing locally. Do you think we should do performance testing between staging and production before merging into develop?

AudreyAltman:
@moltude good suggestion.  I'll look into some options for doing performance testing.

AudreyAltman:
@moltude: I haven't yet found a good way to do automated performance testing, so I did some light ad-hoc testing.  Based on a small set of trials on staging, time spent communicating with the database was reduced by 16% on `source_sets/index`, 7% on `source_set/show`, and 5% on `source/show`.  This should scale favorably in production, where there is more data to process.  

Also, I rebased to develop.

AudreyAltman:
I also took some baseline measurements on production, so we can see how it's performing after this is deployed.

AudreyAltman:
I found a bug in this, so please hold off merging until I can fix it!

AudreyAltman:
@moltude: The bug is fixed, and here’s a summary of all the changes:

It turns out I can’t preload (i.e. include) `tags` in the `source_set` `index` call.  Sometimes, the index needs to display _only sets having specified tags_.  In these cases, the query will _only_ load the specified tags, as opposed to _all_ tags belonging to a set.  This defeats the purpose of the preload, as the view will need to display all of the tags.  This is unfortunate because it means that the app will need to send an additional query for each set in order to get all the tags, but I will just have to try and optimize this with caching.

I did refactor the db call that gets all the tags for the set.  There is a caveat - we really only want tags that are associated with vocabularies for which the value of `filter` is `true`.  Before I was getting all the tags from the database, and then doing some convoluted view logic to get the tags associated with filterable vocabs. I replaced this with model methods.  See https://github.com/dpla/primary-source-sets/pull/165/commits/b78484953b951b4503162a883cf3e84c1f874a79#diff-8aea8a5bfdb4698272ac5b9ee26ba88fR9 and https://github.com/dpla/primary-source-sets/pull/165/commits/b78484953b951b4503162a883cf3e84c1f874a79#diff-8fe92870a7f027a46eed82a51c5c16bcR18  There’s a couple new associated specs tests for the models.  The view logic I replaced is in `app/views/source_sets/_set_tile.html.erb`, and also I no longer need to pass the `filter` variable between view partials.

I also refactored the method `self.with_tags` in `app/models/source_set.rb` so that it’s only one call to the database, rather than one call per tag.  This has the added benefit of causing this method to return an `ActiveModel::Relation` instead of an `Array`, which makes it easier to combine with other methods.  I did, however, need to change the views to work `ActiveModel::Relation` by calling `length` instead of `count` to get the total number of sets returned by the query.

I think that covers the new changes.  I’ve tested it locally, and it seems to be working fine.  I will squash the two commits before merge, but I wanted to leave them separate for the time being for ease of code review.

AudreyAltman:
Just added one last thing - a correction in the `source` `show` view so that it only shows filterable tags, as opposed to all tags.  Tested locally.

AudreyAltman:
ER-Diagram for reference: https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Relational+Database+Model

AudreyAltman:
@moltude squashed, rebased, and ready to merge

moltude:
👍

PR #166, opened 2016-05-18
Add support for cache_store sessions

markbreedlove:
* Add the ability to use the rails `cache_store` session handler.
* Set up the Rails `cache_store`.
* Add the `dali` gem to support memcached.

Is this a good way to set up the Rails cache_store, for other use in addition to sessions?

Is there some better way to set `config.cache_store` that doesn't duplicate code between `development.rb` and `production.rb`? I wasn't able to make it work by creating my own initializer.

The `memcached` package is added to the VM, so you can run `vagrant provision` on it and it will be started. You can then update your `settings.local.yml` with a host `localhost` if you want to use memcached. For example:
```
cache:
  store: mem_cache_store
  opts:
    - localhost
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ gem 'access-granted', '~> 1.0.0'
 gem 'rake', '< 11.0'
 gem 'DPLibrary', git: 'https://github.com/phereford/DPLibrary.git',
                  branch: 'master'
+gem 'dalli', '~> 2.0'
 
 group :test, :development do
   gem 'rspec-core', '~> 3.3.2'
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -47,4 +47,10 @@ Rails.application.configure do
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # Configure cache store.
+  # YAML Hash values (see memory_store example in settings.yml) get turned into
+  # Config::Options objects, but we need them as hashes.
+  cache_settings = Settings.cache.to_hash
+  config.cache_store = cache_settings[:store].to_sym, *cache_settings[:opts]
 end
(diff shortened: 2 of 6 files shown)


markbreedlove:
Ticket: https://issues.dp.la/issues/8399

markbreedlove:
This also modifies the session cookie path, so it doesn't get sent for every request on the rest of the site outside of Primary Source Sets.

AudreyAltman:
I think this looks great.  It appears this will work with fragment caching, which is what I was planning to try.  I don't know of a way to avoid duplicate code in `production.rb` and `development.rb`.  My only question is if you're planning to share the cache between multiple applications, and if so, should we add a namespace? (see http://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-store).

markbreedlove:
Yes, thanks for the reminder about the namespace. I will add that.

markbreedlove:
By the way, the weakness of a cache-based session is that users will obviously be logged out when the cache is cleared. We may want to transition to an ActiveRecord cache at some point. This is just slightly easier to set up and also sets things up for view caching.

markbreedlove:
Ok commit https://github.com/dpla/primary-source-sets/pull/166/commits/19607ec89866b3637cba2967ac2b38d7328df32e adds the namespace and also simplifies the initializer.

I can verify that the namespace is being added:
```
vagrant@pss:/vagrant$ echo "stats cachedump 6 10" | nc localhost 11211
ITEM pss:_session_id:8ade1faff6ffc39410c5a814087e2dda [160 b; 1463589745 s]
END
```

markbreedlove:
oh, hold on, I need to add a documentation comment ...

markbreedlove:
That's better. I'll need to squash all of that before it's merged.

AudreyAltman:
:+1: I'll :merge: as soon as its :squash:ed

PR #168, opened 2016-05-27
Feature/view caching

AudreyAltman:
This adds view fragment caching for set tiles in order to decrease page load time.  It makes use of [cache digests](https://github.com/rails/cache_digests), which comes with Rails 4.  Set tiles appear on `source_sets#index` and `source_sets#show` (in the "Related primary source sets" carousel").  They look like this:

screen shot 2016-05-27 at 2 42 06 pm

Logs show that rendering these tiles greatly slows page load time, and because of the nature of the required database calls, they cannot be effectively cached in the controller.  However, they are a good candidate for fragment caching, especially considering that they are relatively static - the associated image and tags change very infrequently.  This has been tested locally, and significantly reduces page load time.

Changes in this PR include:

* In the views, template dependencies are articulated according the cache digests conventions.  While this particular PR does not make use of template dependencies, it is now set correctly for any future fragment caching.  You can double-check this in your console by running, for example: `rake cache_digests:nested_dependencies TEMPLATE=source_sets/index`.

* Caching is turned on in the development environment for testing purposes (see `config/environments/development.rb`).

* Since the set tiles conceptually represent a set, the cache key for each set tile fragment is tied to the sets table; whenever the sets table is touched, the key expires.  The tricky part is that correctly rendering a set tile depends on data not only from the sets table, but also from sources, images, tags, and vocabularies.  So we need to make sure that when an entry in any one of these tables is updated or destroyed, or when an association between any of these tables is added or removed, the sets table id also updated.   That's what all the changes in the models are about.  It's easy in the `Source` model because a Rails method already exists to take of all these dependencies for `belong_to` relationships - but for all other types of relationships, we have to code it out.

* JavaScript is used to to hide the tags for the set tiles in the `source_sets#show` "Related primary source sets" carousel".  This is so that the same cache fragments can be used in both contexts.  The carousel already depends on JavaScript to render, and defaults to a simple list of sets (as opposed to set tiles) if JavaScript is disabled.  

* Extensive tests to ensure expected behavior.

In the future, I would like to refactor this to eliminate redundancies and make into something more generally useful in different contexts.  I would also like to add one or more fragment caches in other parts of the application.  But for now it might be worth seeing how this behaves in production environment and iterating later.

### DIFF
--- a/app/controllers/source_sets_controller.rb
+++ b/app/controllers/source_sets_controller.rb
@@ -11,10 +11,8 @@ class SourceSetsController < ApplicationController
   def index
     @tags = get_tags_from_params
     @order = params[:order]
-    @published_sets = SourceSet.includes(:small_images).published
-                               .order_by(@order).with_tags(@tags)
-    @unpublished_sets = SourceSet.includes(:small_images).unpublished
-                                 .order_by(@order).with_tags(@tags)
+    @published_sets = SourceSet.published.order_by(@order).with_tags(@tags)
+    @unpublished_sets = SourceSet.unpublished.order_by(@order).with_tags(@tags)
   end
 
   def show
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -1,5 +1,5 @@
 class Source < ActiveRecord::Base
-  belongs_to :source_set
+  belongs_to :source_set, touch: true
   has_many :guides, through: :source_set
   has_many :attachments, dependent: :destroy
 
@@ -17,22 +17,30 @@ class Source < ActiveRecord::Base
 
   has_many :images, through: :attachments,
                     source: :asset,
-                    source_type: 'Image'
+                    source_type: 'Image',
+                    after_add: :touch_self,
+                    before_remove: :touch_self
 
   has_many :large_images, -> { where size: 'large' },
                           through: :attachments,
                           source: :asset,
-                          source_type: 'Image'
+                          source_type: 'Image',
+                          after_add: :touch_self,
+                          before_remove: :touch_self
 
   has_many :small_images, -> { where size: 'small' },
                           through: :attachments,
                           source: :asset,
-                          source_type: 'Image'
+                          source_type: 'Image',
+                          after_add: :touch_self,
+                          before_remove: :touch_self
 
   has_many :thumbnails, -> { where size: 'thumbnail' },
                         through: :attachments,
                         source: :asset,
-                        source_type: 'Image'
+                        source_type: 'Image',
+                        after_add: :touch_self,
+                        before_remove: :touch_self
 
   validates :aggregation, presence: true, 
                           format: { with: /\A[a-zA-Z0-9]+\z/,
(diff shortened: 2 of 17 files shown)


AudreyAltman:
Here's the ER diagram, for reference: https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Relational+Database+Model

app/models/source.rb
```diff
@@ -17,22 +17,30 @@ class Source < ActiveRecord::Base
...
+                          after_add: :touch_self,
+                          before_remove: :touch_self
 
   has_many :thumbnails, -> { where size: 'thumbnail' },
                         through: :attachments,
                         source: :asset,
-                        source_type: 'Image'
+                        source_type: 'Image',
+                        after_add: :touch_self,
+                        before_remove: :touch_self
```
AudreyAltman: It's not ideal here to be re-asserting the `after_add` and `before_remove` statements for `images`, `large_images`, `small_images`, and `thumbnails`.  Really, large_image, small_image, and thumbnail should be refactored to be model methods instead of relations, for example:

     def thumbnail
        images.where(size: 'thumbnail')
      end

That way the `after_add` and `before_remove` would only need to be asserted for `images`.  But that is out of scope for this PR.

app/controllers/source_sets_controller.rb
```diff
@@ -11,10 +11,8 @@ class SourceSetsController < ApplicationController
   def index
     @tags = get_tags_from_params
     @order = params[:order]
-    @published_sets = SourceSet.includes(:small_images).published
-                               .order_by(@order).with_tags(@tags)
-    @unpublished_sets = SourceSet.includes(:small_images).unpublished
-                                 .order_by(@order).with_tags(@tags)
+    @published_sets = SourceSet.published.order_by(@order).with_tags(@tags)
+    @unpublished_sets = SourceSet.unpublished.order_by(@order).with_tags(@tags)
```
AudreyAltman: Before I was preloading all the images needed to render the set tiles with an `includes` statement.  Now, since the set tiles (and their associated images) are cached, we don't need to preload.

app/models/source.rb
```diff
@@ -1,5 +1,5 @@
 class Source < ActiveRecord::Base
-  belongs_to :source_set
+  belongs_to :source_set, touch: true
```
AudreyAltman: For reference on `touch: true` option, see http://guides.rubyonrails.org/association_basics.html#belongs-to-association-reference, section 4.1.2.8

markbreedlove:
I'm finally reviewing this. Sorry for the delay.

markbreedlove:
This looks consistent with the Rails documentation recommends. I think it's OK, and will pass it along for review. (Though see below)

The Javascript that hides the tag HTML elements is the part I'm least enthusiastic about.  Could there be separate view fragments for the carousel items?  We shouldn't use Javascript to fix problems with the view rendering what's basically the wrong HTML.

If that can't be pulled off, it seems that the tags `ul` element should default to `visibility: hidden;`, and the Javascript in the Set page should set `visibility: visible;`, to avoid a jitter in the page layout when it displays and then is hidden.

I agree with using Rails's builtin caching, at least for the time being, but it's not ideal. The cache digests are an OK idea in order to make this feasible for a wide variety of applications without a lot of engineering going into them to devise schemes specifically tailored to each application; but I think more could be done to issue fewer `memcached` requests. We'll have to let that be for now.

AudreyAltman:
@markbreedlove thanks for your review, and for bringing up the use of javascript to hide HTML elements.  I don't know why I didn't realize initially that I could just hide that block with CSS, no javascript necessary.  I made the change here https://github.com/dpla/primary-source-sets/pull/168/commits/02419d13bc8a9565bd4243f6ae3117171cfb1c1f

While I was at it, I also deleted the lines:

   ```.set .related-sets {
      margin: 2em 0;
   }```

because I realized they were superfluous.

markbreedlove:
Thanks, @AudreyAltman. I don't think I have anything more to add and would be happy merging this. Is anyone else in line to review it?  I'll wait for it to be squashed and for your go-ahead before merging.

AudreyAltman:
I don't think anyone else needs to review it.  I'll go ahead and squash.

AudreyAltman:
@markbreedlove okay, squashed and ready to merge!

PR #169, opened 2016-06-01
order filters_for_sets by tag_sequences.position

AudreyAltman:
This fixes a bug introduced in PR#165.  On source_sets#index, the tags listed in the dropdown menu were out of order.  They are supposed to be ordered according to the "position" field in the "tag_sequences" table.  This PR fixes the dropdown menus this so that the positions defined in the tag_sequences table are enforced.

This has been tested locally and is working as expected.  It addresses [ticket #8424](https://issues.dp.la/issues/8424).


no-reply:
Looks good. Can see why the regression got in with optimization, and the fix looks fine to me.

Going to wait on CI and :merge:.

PR #170, opened 2016-06-02
fix order by db call for postgresql

AudreyAltman:
This fixes a bug introduced in [PR#169](https://github.com/dpla/primary-source-sets/pull/169/files).  #169 was tested and worked locally (sqlite database), but failed when deployed to staging (postgresql database).  Postgresql requires that any columns in a ORDER BY clause also appear in a SELECT clause.

This has been tested on staging.  It addresses [ticket #8242](https://issues.dp.la/issues/8424).

### DIFF
--- a/app/helpers/source_sets_helper.rb
+++ b/app/helpers/source_sets_helper.rb
@@ -28,7 +28,8 @@ module SourceSetsHelper
 
       # Get all tags associated with the vocab.
       # Ignore tags that aren't associated with any of the source_sets.
-      tag_list = Tag.joins(:source_sets, :vocabularies)
+      tag_list = Tag.select('tags.*, tag_sequences.position')
+                    .joins(:source_sets, :vocabularies)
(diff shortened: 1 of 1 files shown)


no-reply:
@AudreyAltman if I'm understanding correctly, the previous version failed with errors from PSQL, routed through `ActiveRecord`?

It's a minor frustration that AR doesn't handle this for us, but this fix seems okay. 

@dpla/tech-staff should note that PRs that may affect database queries need to be tested beyond regular CI, to account for differences like this. We could also look into ways to loop PSQL into travis builds, keeping SQLite on the development enviornment.

AudreyAltman:
@no-reply Yes, you are correct about the error.  The error message was: 

`ActiveRecord::StatementInvalid: PG::Error: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list`

Based on a quick search of Rails tickets, it seems that folks are aware of this problem, but are settling with work-arounds for the time being.

PR #174, opened 2016-08-10
Track click through events with Google Analytics

AudreyAltman:
A "click-through" event is when a user clicks on an external link to a provider's website.  This PR adds code to track such an event with Google Analytics.

This PR also refactors existing code related to Google Analytics.  Before this PR, the app was tracking general pageviews, along with tracking views of /source pages as customized events.  The PR brings the code in line with similar code on the frontend: It makes use of the `google-analytics-rails` gem; removes JavaScript logic from views; and uses the data-* attribute to make relevant metadata available to JavaScript from views.

A similar PR was recently approved on the frontend: https://github.com/dpla/frontend/pull/59

This addresses [ticket #8518](https://issues.dp.la/issues/8518).  It has been tested on staging using the Chrome Google Analytics Debugger plugin.


moltude:
As much as I understand JS I think I see what is going on and it looks good. So 👍  from me but not sure if @markbreedlove should give it a pass before merging.

app/assets/javascripts/google-analytics-events.js
```diff
@@ -0,0 +1,67 @@
+/*
+ * Tracks custom events with Google Analytics.
+ * This application uses the google-analytics-rails gem to track general usage
+ * data. This file defines additional user interations to be tracked with Google
+ * Analtyics.
+ * See https://developers.google.com/analytics/devguides/collection/analyticsjs/events
+ */
+
+$(function(){
```
markbreedlove: This syntax is shorthand for `$(document).ready(function() { ... });` so the `$(document).ready()` on line 12 can be removed.

config/environments/development.rb
```diff
@@ -53,4 +53,6 @@
   # Config::Options objects, but we need them as hashes.
   cache_settings = Settings.cache.to_hash
   config.cache_store = cache_settings[:store].to_sym, *cache_settings[:opts]
+
+  GA.tracker = Settings.googleanalytics.tracker
```
markbreedlove: Elsewhere there are references to a `GoogleAnalytics` module. Can you clarify where the `GA` module is defined?

app/assets/javascripts/google-analytics-events.js
```diff
@@ -0,0 +1,67 @@
...
+   * It is set by the google analtyics rails gem if a valid tracker is present.
+   */
+  if (typeof ga !== 'undefined') {
+
+    var item = $(obj).parents("[data-item-id]");
+    var category = "Click Through : " + item.attr("data-provider");
+    var action = item.attr("data-data-provider");
+    var label = item.attr("data-item-id") + " : " + item.attr("data-title");
+
+    ga('send', 'event', category, action, label, {'transport': 'beacon'});
```
markbreedlove: Can you clarify why the `beacon` transport is used here, but not in `trackSourceViewEvent`?
Should the transport be set at a higher level?
https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits

AudreyAltman: @markbreedlove I had originally used the `beacon` transport b/c pages were being unloaded when users clicked through to provider pages (ie. the links were opening in the same tab).  The `beacon` helps ensure that the data gets sent to Google Analytics on page unload.  However, part of this PR changes the links so that they [open in new tabs](https://github.com/dpla/primary-source-sets/pull/174/files#diff-b6a20150fda7fa0a5aef93e64f6a5c28R74).  This makes it un-necessary to use `beacon`, so I will remove it and just use the default transport method instead.

app/assets/javascripts/google-analytics-events.js
AudreyAltman: Good catch!

config/environments/development.rb
AudreyAltman: Turn out `GA` is an alias for `GoogleAnalytics`.  I'll use `GoogleAnalytics` instead of `GA` throughout for clarity, and make a note in the config files that it's defined in the `google-analytics-rails` gem.

markbreedlove:
I'm fine with merging this, if there are no other comments...

no-reply:
:+1:

PR #177, opened 2016-08-25
pin all gems to tiny versions

AudreyAltman:
Pin all gems to the tiny version that are currently deployed to production.


no-reply:
:+1:

PR #179, opened 2016-11-02
configure x-frame-options

AudreyAltman:
This allows us to set X-Frame-Options for primary source sets.  Along with accompanying changes to [automation](https://github.com/dpla/automation/pull/205) and [aws](https://github.com/dpla/aws/pull/47), it will allow OER commons to display an iframe of primary source sets.  OER commons indexes our sets in their educational resources discovery tool.

I have run a local deployment with the aws and automation changes, and verified with curl:

screen shot 2016-11-02 at 9 22 09 am

This addresses [ticket #8490](https://issues.dp.la/issues/8490).

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -23,5 +23,10 @@ module PrimarySourceSets
 
     # Autoload contents of lib
     config.autoload_paths += %W(#{config.root}/lib/primary-source-sets)
+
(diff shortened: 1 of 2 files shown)


markbreedlove:
Great, thanks. This looks good to me.

PR #186, opened 2017-06-12
add json view for sets#index

AudreyAltman:
This adds a JSON view for the `sets#index` page.  It lists the URIs of all the sets, which will allow us to harvest sets via HTTP.  It also lists the name and schema.org type for each set, which follows precedent set elsewhere in the app (ie. https://github.com/dpla/primary-source-sets/blob/develop/app/views/source_sets/_show.json.erb#L57:L80).

This is the first step to addressing [ticket DT-1393](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1393).  It has been tested locally and on staging.  It is currently deployed to staging, in case anyone wants to use it to test the primary source sets harvester.

The second step to addressing ticket DT-1393 is [ingestion3#17](https://github.com/dpla/ingestion3/pull/17).

### DIFF
--- a/app/controllers/source_sets_controller.rb
+++ b/app/controllers/source_sets_controller.rb
@@ -13,6 +13,16 @@ class SourceSetsController < ApplicationController
     @order = params[:order]
     @published_sets = SourceSet.published.order_by(@order).with_tags(@tags)
     @unpublished_sets = SourceSet.unpublished.order_by(@order).with_tags(@tags)
+
+    ##
+    # Render HTML or JSON formats unless controller has already redirected or
+    # rendered (ie. in the check_login_and_authorize method).
+    unless performed?
+      respond_to do |format|
+        format.html
+        format.json { render partial: 'source_sets/index.json.erb' }
+      end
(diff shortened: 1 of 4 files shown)


markbreedlove:
Looks good. Looks like it's completing the JSON functionality that was partially added in April 2016.

Are you OK with me merging this as it is now?

markbreedlove:
Actually, I've just found one thing that might want to be changed. The `thumbnailUrl` field returns a `cloudfront.net` URL that does not work, e.g. `//d1dbpjegy13z6i.cloudfront.net/northern-draft-riots-civil-war_01_649f367dc5b658216d83844037d7b4f3_sm.jpg`. I'm checking to see whether this is a general problem or just something that's misconfigured on my end with AWS.

markbreedlove:
Yes, I've seen this CloudFront issue before. Because we have a CloudFront distribution that's "aliased" in Route53 to a DNS A record, you have to refer to it with the A name. I'll contact you on another channel.

PR #189, opened 2017-08-24
Dt 1550 postlight tag lists

AudreyAltman:
This adds tag lists to the JSON API, for the purpose of creating dropdown "Subject" and "Time period" menus.


AudreyAltman:
I'm guessing all these merges are showing up here b/c of the recent change in the [Software Release Process](https://digitalpubliclibraryofamerica.atlassian.net/wiki/spaces/TECH/pages/16121895/Software+release+process).  We used to merge with `no-ff` but that was taken out.  Could that be the cause?

moltude:
@AudreyAltman I think so

PR #193, opened 2017-09-20
handle array of provider URLs

AudreyAltman:
This handles an array of provider URLs coming through the DPLA API.  Currently, Digital Maine's provider URLs are in array format.  This has been tested on staging.

### DIFF
--- a/app/models/dpla_item.rb
+++ b/app/models/dpla_item.rb
@@ -34,7 +34,7 @@ class DplaItem < ActiveModelBase
   end
 
   def digital_resource_url
-    item.try(:url)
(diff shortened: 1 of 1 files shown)


AudreyAltman:
Good question, and yes, we do need the `flatten`.  It may be an Array, or just a String.  If it's already an Array, than `Array()` will make an array of an array, hence the need for `flatten`.

markbreedlove:
From my irb console:
```
irb(main):005:0> Array(["x"])
=> ["x"]
```

AudreyAltman:
Ha!  I was wrong, you were right.  I'll fix it.

PR #194, opened 2017-09-28
Kludge in "rep" thumbnail images for Postlight frontend

markbreedlove:
Add "repImageUrl" properties alongside existing "thumbnailUrl" properties in the API.  These new properties are file names that are the same as the thumbnail images, except that they have "_thumb" or "_sm" torn off, if they are present, and the basename ends with "_rep".

See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1615

I've solicited feedback from Postlight about this, but maybe in the meantime this can be reviewed so it's ready to go if they think it's OK.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -76,4 +76,9 @@ module ApplicationHelper
     "via=#{Settings.twitter_username}&" \
     "text=#{content_for :title}"
   end
+
+  def rep_file_name(filename)
+    filename.sub(/(?:_sm|_thumb)?(\.[a-z]*)$/, '_rep\1')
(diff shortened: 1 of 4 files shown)


markbreedlove:
@fdsimms Are you able to comment on this PR?

fdsimms:
this sounds good to me

markbreedlove:
Thanks, @fdsimms !

@AudreyAltman and @mdellabitta  thanks for your reviews. I've made some minor fixups, which you can see above. I'm assuming they won't affect your reviews, and I'll merge this a.s.a.p. and try to get it deployed.

PR #196, opened 2017-11-01
make more json fields html safe

AudreyAltman:
This adds more HTML safe fields to the JSON API.  It has been tested locally.


AudreyAltman:
This is failing to do an issue with Jasmine.  I have made a ticket to address the problem: https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1832  In the meantime, I am merging this so as not to block related work.

PR #199, opened 2017-11-01
change bundler without syntax

AudreyAltman:
This fixes syntax to align with a change in Bundler 1.6 (see https://github.com/bundler/bundler/commit/8ccf49c3b4a086f1e0cad2130dc30c04304175b7#diff-75219bd6bc0defc0e7f4ba322d92f950L198).  This fixes deployment to staging, which was failing.


AudreyAltman:
😆

